PR #7, opened 2020-04-06
Initial backend setup

tjanowsk:
Initial spring setup and some dummy GET endpoints for now


src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+
```
QuarkNerd, line 1: is this file needed?

tjanowsk: According to https://www.tutorialspoint.com/spring_boot/spring_boot_application_properties.htm
this file contains spring's variables, which we're not using at the moment.
Rather than removing this, I'm tempted to add `server.port` and `spring.application.name` to make its purpose more obvious,

floyd924: Some of the things you might put in here soon are the database connection details, or any variables you want to set differently between Test, Dev and prod :)

backend/src/main/java/com/rota/api/HelloController.java
```diff
@@ -0,0 +1,43 @@
...
+public class HelloController {
+  private AtomicLong id = new AtomicLong();
+  private String template = "Hello %s!";
+
+  /**
+   * /hello will return "Hello World!".
+   * /hello?message=foo witll return "Hello foo!".
+   * @param message Message to print
+   * @return HelloWorld response DTO
+   */
```
JAStrachan, line 20: A little later and in a separate pull request we can add Swagger integration. It generates a webpage with all our endpoints and info of what's going in and out! https://swagger.io/tools/swagger-ui/ Is what we used on MTG for this. Later I'll try and find some better resources for that.

backend/src/main/java/com/rota/api/dto/HelloWorld.java
```diff
@@ -0,0 +1,11 @@
+package com.rota.api.dto;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
```
JAStrachan, line 6: Nice! I always forget about the cool fancy lombok annotations!

PR #24, opened 2020-04-07
Created base react app with typescript template

AdrianBorg:
ran `npx create-react-app frontend --template redux-typescript` in the base project directory

and added redux logger to middleware

-----------------------------

This should've been in its own PR, but its done here:

* created workflow to install, build and run tests for front end
* changed workflows to run only when a file in their repo is changed
* added a workflow to work around an issue as described in [this link](https://github.community/t5/GitHub-Actions/Github-actions-and-required-checks-in-a-monorepo-using-paths-to/td-p/34829) where required checks go on forever if they are skipped (e.g a front end PR will wait for the backend build check to finish, even though it doesnt run since the changes in the PR are just front end ones)

Once this is merged, we can enable the required checks for setup as shown here:
image


floyd924:
Yep, this looks to me like th boiler plate code we need. PR appears to only be a draft, though?

AdrianBorg:
> Yep, this looks to me like th boiler plate code we need. PR appears to only be a draft, though?

Yep, just getting redux set up too before making it a proper one

.github/workflows/backend.yml
```diff
@@ -1,10 +1,18 @@
...
 
-name: CI
+name: BackendCI
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events but only for the master branch
-on: [push, pull_request]
+on: 
+  push:
+    branches: [master]
```
tjanowsk, line 9: Can we remove branches:[master] in both cases? I think there's value in running CI on push to any branch.
Also, we can't push to master directly.

AdrianBorg: it is now, forgot to add that line

AdrianBorg: true, but should we have the workflow run on every push, even when we dont have a PR up?

tjanowsk: It's free, so I don't see any reasons not to run everything. It would save us from opening PRs from branches with failed builds.
If anything, I would be more inclined on removing checks on PRs because they are necessarily preceded by a push.

AdrianBorg: fair point, I'll switch over to putting htem on just pushes

frontend/src/features/counter/Counter.tsx
```diff
@@ -0,0 +1,60 @@
...
+import {
+  decrement,
+  increment,
+  incrementByAmount,
+  incrementAsync,
+  selectCount,
+} from './counterSlice';
+import styles from './Counter.module.css';
+
+export function Counter() {
```
tjanowsk, line 12: :+1: for using function as opposed to class components

AdrianBorg: tbh this is auto generated, I didnt add thing, i left it in for people who start off to maybe have something to work off of

frontend/src/App.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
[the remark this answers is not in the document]
AdrianBorg: Think its standard if we're going with the ducks file structure

.github/workflows/frontend.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  build-and-test-frontend:
+    runs-on: ubuntu-latest
+    env:
+      working-directory: './frontend'
+
+    # strategy:
+    #   matrix:
+    #     node-version: [12.x] #[8.x, 10.x, 12.x]
+
```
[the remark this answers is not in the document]
AdrianBorg: will do

AdrianBorg:
> Looks good. My only concern is that this PR contains some disposable elements we're never going to use.
> These include all the images (logo192.png, logo512.png, favicon.ico and logo.svg) as well as all the css files. We'll need to clean them up at some point so might as well do it now.

I'm of the mind that we should remove them when they are replaced, that way if someone comes in they've got something to follow.

frontend/src/features/counter/Counter.module.css
```diff
@@ -0,0 +1,74 @@
+.row {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
```
JAStrachan, line 6: This might be the default setting but wanted to ask. Are you wanting to use css modules or do you want to use a preprocessor like SASS? There is merits to both and its ultimately up to you guys but thought to ask!

.github/workflows/backend.yml
```diff
@@ -1,15 +1,18 @@
 # This is a basic workflow to help you get started with Actions
 
-name: CI
+name: BackendCI
 
 # Controls when the action will run. Triggers the workflow on push or pull request
 # events but only for the master branch
```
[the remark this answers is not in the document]
AdrianBorg: True, its on all commits. I'll change it over

README.md
```diff
@@ -1,2 +1,17 @@
...
+
+* Typescript
+* React
+* Redux
+  * Redux Toolkit ([Why use toolkit?](https://redux-toolkit.js.org/introduction/quick-start))
+    * Redux Thunk
+    * Redux Devtools
+  * Redux Logger
+
+
```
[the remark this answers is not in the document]
AdrianBorg: I'll move it over

frontend/src/features/counter/Counter.module.css
AdrianBorg: Yeah this is the default that came out of the cra template, I've never actually used the modules thing. I think I might leave it in and deffer a decision to whenever we get to start styling some components. Plus not sure what the others would be comfortable with using most.

frontend/src/app/store.ts
```diff
@@ -0,0 +1,25 @@
...
+  ]
+
+const reducer: ReducersMapObject = {
+  counter: counterReducer,
+}
+
+export const store = configureStore({
+  reducer,
+  middleware
+});
```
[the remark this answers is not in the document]
AdrianBorg: Good catch, I'll add the line for disabling it in production

PR #42, opened 2020-04-09
Added swagger to the project (see backend/README.md)

tjanowsk:
Used swagger on the server.

Server's attack sharply rose.
Server is confused.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build
 .project
 .settings
 *.iml
+.vscode
--- a/backend/README.md
+++ b/backend/README.md
@@ -25,6 +25,14 @@ ln -s ../../.githooks/pre-commit .git/hooks
 chmod +x .git/hooks/pre-commit
 ```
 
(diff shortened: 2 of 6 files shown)


floyd924:
Would it be worth adding the required annotation to a, existing hello world endpoint so as to show the next developer how to annotate future endpoints?

.gitignore
```diff
@@ -6,3 +6,4 @@ build
 .project
 .settings
 *.iml
+.vscode
```
AdrianBorg, line 9: You're right here in that we probably don't want to add he .vscode dir into out repo. Maybe i should just document what should be included in the readme for PR #43, and also add the .vscode to the gitignore there.

tjanowsk: I added it before looking at your PR. Since it contains only 1 json file I'm not bothered by it.

AdrianBorg: Do we want to remove it from the gitignore, or the repo then?

PR #43, opened 2020-04-09
Feature/4 frontend linting implementation

AdrianBorg:
Added linting to front end stuff.

Used the airbnb-typescript template for eslint rules, we might want to change this. I did have to edit some rules and introduce a few new packages to get eslint, prettier, typescript and the airbnb rules to work together nicely

Added the linting and auto fixes to be done pre-commit using husky

Also added the `vscode` directory in the root folder to point at the linting config files for if people open the project from there (otherwise linting wont work with vscode plugins). Not sure if we want to keep this or not.


frontend/package.json
```diff
@@ -40,5 +40,30 @@
...
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-react": "^7.19.0",
+    "eslint-plugin-react-hooks": "^2.5.1",
+    "husky": "^4.2.3",
+    "lint-staged": "^10.1.3",
+    "prettier": "2.0.4"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
tjanowsk, line 60: Husky does not play well with custom pre-commit .git/hook from the other PR.
It does not install properly if a hook already exists, I had to wipe the .git/hooks directory and reinstall husky to get it to work, but then the back end hook is no longer in place.

Can we just add the linters to the existing .githooks/pre-commit file instead?

frontend/package.json
```diff
@@ -40,5 +40,30 @@
...
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb-typescript": "^7.2.0",
+    "eslint-config-prettier": "^6.10.1",
+    "eslint-plugin-import": "^2.20.2",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-react": "^7.19.0",
+    "eslint-plugin-react-hooks": "^2.5.1",
+    "husky": "^4.2.3",
+    "lint-staged": "^10.1.3",
+    "prettier": "2.0.4"
```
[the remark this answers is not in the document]
AdrianBorg: This was recommended by the prettier themselves. On their [installation page](https://prettier.io/docs/en/install.html) they say:

>We recommend pinning an exact version of prettier in your package.json as we introduce stylistic changes in patch releases.

frontend/package.json
AdrianBorg: Wasn't aware of this, I'll go ahead and start switching over to the github hooks for the pre commit checks.

frontend/.prettierrc
```diff
@@ -0,0 +1,10 @@
+{
+  "trailingComma": "es5",
+  "tabWidth": 2,
+  "semi": true,
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "arrowParens": "always",
```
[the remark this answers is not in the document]
AdrianBorg: Personally, I'd say the opposite, and I think would make it more consistent if we have some arrow functions with multiple arguments (and would make it more friendly to those who aren't as familiar with arrow functions).

But not that bothered about changing it over to "avoid" instead. [Default however is "always"](https://prettier.io/docs/en/options.html#arrow-function-parentheses)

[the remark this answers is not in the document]
tjanowsk: I'm not bothered either, just thought it looked weird.

frontend/package.json
tjanowsk: OK. It's probably a good idea to have a single source of truth for the git hooks anyway.

frontend/package-lock.json
```diff
@@ -6383,6 +6543,193 @@
       "resolved": "https://registry.npmjs.org/https-browserify/-/https-browserify-1.0.0.tgz",
       "integrity": "sha1-7AbBDgo0wPL68Zn3/X/Hj//QPHM="
     },
+    "human-signals": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/human-signals/-/human-signals-1.1.1.tgz",
+      "integrity": "sha512-SEQu7vl8KjNL2eoGBLF3+wAjpsNfA9XMlXAYj/3EdaNfAlxKthD1xjEQfGOUhllCGGJVNY34bRr6lPINhNjyZw==",
+      "dev": true
+    },
+    "husky": {
```
tjanowsk, line 6552: This needs to be updated from package.json

frontend/README.md
```diff
@@ -6,10 +6,24 @@ We are using the ["ducks" file structure](https://github.com/erikras/ducks-modul
...
-    * Redux Devtools
-  * Redux Logger
+- Typescript
+- React
+- Redux
+  - Redux Toolkit ([Why use toolkit?](https://redux-toolkit.js.org/introduction/quick-start))
+    - Redux Thunk
+    - Redux Devtools
+  - Redux Logger
+- Husky
```
tjanowsk, line 16: not used anymore

frontend/package.json
```diff
@@ -24,7 +24,8 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "lint": "npx prettier --write ./src/**/*.{js,jsx,ts,tsx} && npx eslint --fix ./src/**/*.{js,jsx,ts,tsx}"
```
tjanowsk, line 28: I don't think prettier can ever throw an error, so it doesn't need to be a part of CI

AdrianBorg: Good catch!

I've changed the `--write` flag in the prettier command to `--check`. This should run and fail if it finds stuff that prettier could format.

This way we'd make sure that anything committed to the repo should have had prettier run on it first. (And if people have the pre-commit hook set up, that should be done automatically for them)

AdrianBorg:
> I approve but have you checked that this works with the prettier and eslint plugins in vscode

I have, well I've been using vscode with the plugins myself and making sure stuff works as expected along the way. Also got the .vscode dir in the root folder to point to the right stuff (since the plugins would expect to be in the frontend folder)

Issue #36, opened 2020-04-09
Change shape of user database object to include active field


mangan360:
There is a requirement to keep employee shift data once they have been "deleted" from the employee list. Instead of doing a strict delete of employee data there will be a flag set in the user database object "Active". 
It should have a boolean indicating whether a user is still actively working in the company.

Acceptance Criteria: 
- Given a user, when they are working for the company they should have a the Active value set to true in the database. 
- Given a user, when they no longer works for the company then they should have a flag of Active set to false in the database.

PR #44, opened 2020-04-10
Create local database

tjanowsk:
Create a database and populate it with dummy data.
Instruction on how to access it are found in backend/README.md

The db structure is based on https://drive.google.com/open?id=1tZOHUczPERUcGlcnM7P4wwMiSAwdHCuH
HOWEVER I didn't know how to make a list of preferred dates in an SQL database, so I created a separate table for that.


backend/src/main/resources/data.sql
```diff
@@ -0,0 +1,55 @@
...
+INSERT INTO ENGAGEMENT(start, end, type, hours_worked, staff_id)
+VALUES
+    ('2018-08-03 00:00', '2018-08-03 01:00', 'shift', 1, 1),
+    ('2018-08-04 09:00', '2018-08-04 17:00', 'holiday', 10, 1),
+    ('2017-08-03 10:00', '2017-08-03 20:00', 'shift', 10, 2),
+    ('2018-08-03 00:00', '2018-08-13 00:00', 'holiday', 80, 2),
+    ('2018-08-03 00:00', '2018-08-03 01:00', 'shift', 1, 3),
+    ('2018-08-03 00:00', '2018-08-03 01:00', 'holiday', 1, 3),
+    ('2018-08-05 12:00', '2018-08-05 17:00', 'shift', 3, 4),
+    ('2020-08-03 00:00', '2021-08-03 01:00', 'holiday', 1, 4);
```
AdrianBorg, line 55: Not sure if these are being used right now, but could we have these in a separate .sql file to the table setup

Or have the table statements in an `init.sql` file. This way we'd be able to just run the table sql file to initialise a clean db, then run the populating sql file to add data into the tables separately if we wanted to

backend/src/main/resources/data.sql
```diff
@@ -0,0 +1,55 @@
...
+    sunday BOOLEAN
+);
+
+CREATE TABLE ENGAGEMENT (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    start DATETIME NOT NULL,
+    end DATETIME NOT NULL,
+    type ENUM('shift', 'holiday') NOT NULL,
+    hours_worked DECIMAL,
+    staff_id INT REFERENCES STAFF(id) ON DELETE CASCADE
```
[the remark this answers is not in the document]
AdrianBorg: The SO link probably work better with issue #36 as it looks its one of the solutions proposed

backend/README.md
```diff
@@ -25,6 +25,23 @@ ln -s ../../.githooks/pre-commit .git/hooks
...
 ```
 
+## Server endpoints
+
+To list all available endpoints, run the server and go to `{server-root}/swagger-ui.html`.
+For example, when running locally, visit:
+```
+http://localhost:8080/swagger-ui.html
+```
+
```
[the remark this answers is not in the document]
tjanowsk: I merged into master and then tried rebasing this branch.
Clearly something went wrong down the line. I'll try fixing the rebase and force push or something along these lines.

backend/src/main/resources/data.sql
```diff
@@ -0,0 +1,55 @@
+CREATE TABLE STAFF (
+  id INT AUTO_INCREMENT  PRIMARY KEY,
+  first_name VARCHAR(255) NOT NULL,
+  surname VARCHAR(255) NOT NULL,
+  start_date DATE NOT NULL,
+  contracted_hours DECIMAL NOT NULL,
+  hourly_rate DECIMAL NOT NULL,
+  role ENUM('user', 'manager') NOT NULL,
+  job_title VARCHAR(255) NOT NULL
```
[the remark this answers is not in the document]
tjanowsk: I'd keep the surname as well, but I see your point. Will fix.

backend/src/main/resources/data.sql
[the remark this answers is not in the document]
tjanowsk: Have we decided what to do with 'removed' staff? 
One idea on the call was to add an 'active' boolean flag, in which case we'd never remove the user from the database, but I'm not sure this decision has been finalised.

backend/src/main/resources/data.sql
tjanowsk: Agreed. I kept them in one file out of pure laziness. Will fix.

backend/src/main/resources/data.sql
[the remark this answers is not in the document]
tjanowsk: added active flag since it's a one-liner, I'll link the other issue to this PR as well

backend/src/main/resources/schema.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE STAFF (
```
floyd924, line 1: Sorry I havent had time to look at this until now.

So i mentioned in the call that I'm happy for you guys to move on doing it like this if you have an idea of how it works. However, I'm more familiar using classes and JPA repositories with a Sping project.

Here you've outlined the tables that we need. What you should also do in this PR is create a java class for each of these tables you're creating.

Then, rather than using a db maipulation class in the other PR where you're manually writing the SQL, we can create a repository class, which comes with default methods such as `staffRepository.delete(id)`, or `staffRepository.add(newStaff)`

For example, on the magic app we have a sql table called UserBalance, which looks like this:
Screen Shot 2020-04-14 at 13 12 56


And to store this, we've created a repository in the same directory which looks like this:
Screen Shot 2020-04-14 at 13 13 11


Then in all you classes where you need to use this repository (eg from an endpoint) you can use `@Autowired StaffRespository staffRepository` at the top of the class and then just call those staffRepository.getAll() or staffRepository.save(object) methods which come predefined. If you need custom methods for the repo you can define those in the Repository class too

backend/src/main/resources/schema.sql
```diff
@@ -0,0 +1,33 @@
...
+    staff_id INT REFERENCES STAFF(id) ON DELETE CASCADE,
+    PRIMARY KEY(staff_id),
+    monday BOOLEAN,
+    tuesday BOOLEAN,
+    wednesday BOOLEAN,
+    thursday BOOLEAN,
+    friday BOOLEAN,
+    saturday BOOLEAN,
+    sunday BOOLEAN
+);
```
JAStrachan, line 23: I think this would be better modelled for the now as a array of VARCHARs. It can still be modelled as a Enum in the java code and thrown into the database using the the Enums toString method. https://www.baeldung.com/jpa-persisting-enums-in-jpa Helps though admittedly it is using JPA instead of JBDC.

tjanowsk: Not sure how to create an array of varchar in sql (or an array of anything without creating another table).
Writing a String[] in staff.java and having jpa handle it results in varbinary(255), which is not something we can set easily with data.sql. It also generates spotbugs security warnings.

backend/src/main/java/com/rota/database/orm/engagement/Engagement.java
```diff
@@ -0,0 +1,34 @@
...
+
+@Data
+@Builder
+@Entity
+@AllArgsConstructor
+@NoArgsConstructor
+public class Engagement {
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  int id;
```
floyd924, line 26: To make this easier for someone unfamiliar with the codebase to read I would add an empty line between each field (but not between an annotation and the field it is related to)

backend/src/main/java/com/rota/database/orm/engagement/Engagement.java
```diff
@@ -0,0 +1,34 @@
...
+  Instant start;
+  Instant end;
+  @Enumerated(value = EnumType.STRING)
+  EngagementType type;
+  double hoursWorked;
+  @ManyToOne(fetch = FetchType.LAZY)
+  Staff staff;
```
floyd924, line 33: I also notice you've put `private` on fields in the Staff class, but not here. Can you add them here as well?

backend/src/main/java/com/rota/database/orm/staff/Staff.java
```diff
@@ -0,0 +1,34 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class Staff {
+  @Id
+  @GeneratedValue(strategy = GenerationType.IDENTITY)
+  private int id;
+  private String firstName;
+  private String surname;
+  private boolean active;
```
floyd924, line 26: same as above, I would add an empty line between each of these to make it more readable

backend/src/test/java/com/rota/database/DbHandlerTests.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
+import org.springframework.test.annotation.DirtiesContext;
+
+@EnableJpaRepositories
+@DataJpaTest
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
floyd924, line 21: If this is how we get around having a pre-populated database before tests, that's okay for now.

In the apps I've built we have 2 different directories which create the sql tables. 1 used for prod/tests, and 1 used for dev. Only the Dev one contains seed data, which keeps it away from the tests. The downside is that if we make changes to 1 script we have to change it in 2 places, so it's not a perfect solution.

backend/src/main/java/com/rota/database/orm/engagement/Engagement.java
tjanowsk: Yes, I used to have this as `@Value` instead of `@Data`, which automatically labels all fields as `private final`, but had to convert back to `@Data` to have an empty constructor (required by jpa). I forgot to put `private` back after that.

tjanowsk:
> Looks good to me.
> 
> Just one question to check I've understood correctly, were we able to get rid of the `schema.sql` script because the table structures are set up in the java classes?

Right, jpa construct the database automatically from the dto objects (@Entity annotation does this, I believe), so there is no reason for schema.sql anymore.

PR #47, opened 2020-04-13
feature/46 add database manipulation class

tjanowsk:
TO DO:
- at the moment if staff table is changed we need to modify 3 files to reflect that, think how to make the code more maintainable by improving this


backend/src/main/java/com/rota/database/DbHandler.java
```diff
@@ -0,0 +1,254 @@
...
+  JdbcTemplate jdbcTemplate;
+
+  /**
+   * Erases all database entries.
+   */
+  public void wipeDb() {
+    jdbcTemplate.update(
+        "DELETE FROM STAFF WHERE TRUE;"
+            + "DELETE FROM ENGAGEMENT WHERE TRUE;"
+            + "DELETE FROM PREFERRED_DATES WHERE TRUE;"
```
QuarkNerd, line 35: I believe this is faster `DELETE * FROM table_name;`  
And it also feels like better practice then checking if true is true

QuarkNerd:
Hey, so just a couple of questions. (Please note im not trained in Java. but C#.

Firstly why do we remove IDs from things before we add them to the table? if its the tables job to make them, then why do we add one in the first place? And should we not add a uniwue identifier we can have from our end (email or uid from auth)?

Secondly, have you considered using a library to make SQL string making easier to do and read https://stackoverflow.com/questions/5620985/is-there-any-good-dynamic-sql-builder-library-in-java? And if you don't do this, please consider moving SQL strings in get and remove functions to separate reusable functions.

Unfortunately, I can't help with your specific problem because I'm bot yet sure what everything does.

tjanowsk:
> Firstly why do we remove IDs from things before we add them to the table? if its the tables job to make them, then why do we add one in the first place? And should we not add a uniwue identifier we can have from our end (email or uid from auth)?
> 

The database increments the id automatically. We'll probably want to add some external identifier, but it wasn't part of the current design.
Now that you mention it, it's probably easier to remove id from getPropertyMap of the object. I wanted to keep the mapping faithful originally, but since it's only used for writing to db (the mapper classes are responsible for reading) it's probably better to not include it i the first place.

> Secondly, have you considered using a library to make SQL string making easier to do and read https://stackoverflow.com/questions/5620985/is-there-any-good-dynamic-sql-builder-library-in-java? And if you don't do this, please consider moving SQL strings in get and remove functions to separate reusable functions.
Since we're using spring framework I just went with spring's native database support without looking too hard for alternatives. The good news is that we can use spring annotations to sanitize the unit tests.
I'll look into the alternatives.

JAStrachan:
So the database should create the ids. We can create the object in the java code and when we "save" it to the database it creates the id for us and returns the object with that unique identifier. This "save" is admittedly a JPA term.

JAStrachan:
Though the one place this is subjective is the person table. We could use the uid from the auth as our identifer but I know some projects have databases have a unique key the database creates and the unique string the auth service creates in there "user" tables.

JAStrachan:
I would defintely have a look at JPAs for spring as they can be quite easy to use. In certain cases it can feel slightly magic but you can get into a quite a good workflow. https://www.baeldung.com/the-persistence-layer-with-spring-data-jpa this is a fairly decent introduction and will lead to other relevant articles.

tjanowsk:
Using jpa instead

PR #54, opened 2020-04-16
feature/33 Create get myShifts endpoint

tjanowsk:
Created /myShifts endpoint which returns user's shifts when queried with an 'authentication token' in the header.
The authentication is not yet implemented, so instead we simply pass user ID inside the 'Authorization' header field.

The endpoint accepts 2 optional arguments which are 'start' and 'end'. The returned engagements will have to start and finish between these two time stamps.

Verify with: 
curl http://localhost:8080/myShifts -H 'Authorization: 1'


backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -0,0 +1,75 @@
...
+    return engagementRepository.findByStaffId(staffId).stream()
+        .filter(engagement ->
+            engagement.getStart().isAfter(start)
+          && engagement.getEnd().isBefore(end))
+        .map(EngagementDto::fromEngagement)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Returns all shifts of a given staff ember.
```
floyd924, line 37: typo in staff member

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.server.ResponseStatusException;
+
+@RestController
+@Api(value = "Manages the responses to staff-related queries")
+public class StaffController {
+  @Autowired
+  EngagementRepository engagementRepository;
+
+  private List<EngagementDto> getStaffEngagementsBetween(int staffId, Instant start, Instant end) {
```
floyd924, line 27: Once we get more methods like this which are used by the get and post endpoints we could probably move them into a `StaffManager` class, or something similar, so as to only have endpoints in this file, keeping it tidy

emab: In my previous project we used a `StaffService` class to handle any business logic which was then much easier to test.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -0,0 +1,75 @@
...
+          Instant start,
+
+      @RequestParam(name = "end", required = false)
+      @ApiParam(value = "end time")
+          Instant end
+  ) {
+    final Optional<String> user = AuthenticationUtils.getUserFromToken(authString);
+    user.orElseThrow(
+        () -> new ResponseStatusException(HttpStatus.FORBIDDEN, "Failed to authorize.")
+    );
```
JAStrachan, line 61: I believe you can just write .orElseThrow(new Response...)

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -0,0 +1,75 @@
...
+
+    int staffId;
+    try {
+      staffId = Integer.parseInt(user.get());
+    } catch (NumberFormatException e) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid user ID format in the "
+          + "token.");
+    }
```
JAStrachan, line 69: This isn't all of it but could you extract all the authentication code to it's own static class? Will hopefully make it easier to refactor later as we just have to change the innards or that method

backend/src/main/java/com/rota/api/dto/EngagementDto.java
```diff
@@ -0,0 +1,38 @@
...
+  int staffId;
+
+  @ApiModelProperty(value = "Starting time of an engagement")
+  Instant start;
+
+  @ApiModelProperty(value = "End time of an engagement")
+  Instant end;
+
+  @ApiModelProperty(value = "Type of engagement, e.g. SHIFT, HOLIDAY etc.")
+  EngagementType type;
```
JAStrachan, line 23: These model annotations are good! They are probably the ones we want. But there isn't too much point in using them for swagger unless we use them to add all the info they allow! But that can wait till another pull request when we have a firmer understanding of what we want exactly from each endpoint. Between `@ApiParam` and `@ApiModelProperty` https://www.baeldung.com/swagger-apiparam-vs-apimodelproperty

backend/src/main/java/com/rota/api/StaffController.java
JAStrachan: Ah I see you already have the authentication class! I would probably still move the extraction of the token to it! Let it deal with all the exceptions!

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: orElseThrow takes a Supplier as an argument (i.e. a void function), which is incomplatible with feeding it an exception directly.
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseThrow-java.util.function.Supplier-

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: I need to throw the runtime exceptions inside the controller class to get meaningful feedback on the front end.
This is why I still kept some optional juggling.

Unfortunately java does not have Either object, it would be much nicer to work with Either(Error, Value)

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -0,0 +1,59 @@
...
+  ) {
+    if (!AuthenticationUtils.validateToken(authString)) {
+      throw new ResponseStatusException(HttpStatus.FORBIDDEN, "Failed to authorize.");
+    }
+
+    final Optional<Integer> user = AuthenticationUtils.getUserFromToken(authString);
+    user.orElseThrow(
+        () -> new ResponseStatusException(HttpStatus.BAD_REQUEST, "Invalid user ID format in the "
+            + "token.")
+    );
```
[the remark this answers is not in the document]
tjanowsk: Why are all those https status codes so similar? :scream:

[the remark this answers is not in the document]
floyd924: On my last client project we stuck with bad_request if we weren't sure, as long as there was a descriptive warning message

backend/README.md
```diff
@@ -20,6 +20,18 @@ For example, when running locally, visit:
 http://localhost:8080/swagger-ui.html
 ```
 
+### myShifts endpoint
+
+Returns shifts of a given user (inferred from the authentication token).
+At the moment authentication has not been implemented yet, so the "token" is just user ID.
+Example call:
+` curl http://localhost:8080/myShifts?start=1410-07-15T00:00:00Z&2100-01-01T00:00:00Z -H
+ 'Authorization: 1' `
```
QuarkNerd, line 29: should there be an `end=` in the request URL? (after `&`).  also when I run the command I get `bash: -H command not found`

tjanowsk: Yes, there should be an `end=`
The error is weird, -H is a curl flag not a bash command. Are you running everything on one line?

QuarkNerd: Yup 
image

QuarkNerd: Solved it, need quotes around the request, probably dependent on the shell

Issue #49, opened 2020-04-16
Create basic Home and Sidebar components


floyd924:
Create a "Sidebar" react component which contains 3 rectangular buttons labelled as per designs. Add a SCSS file for this component, import it, and set rough heights/widths for these divs. Do not worry about colour or shadow yet.

Create a "Home" component which will render 2 containers. Make the first container contain the "Sidebar" component and set the width to 25% of the page. Make the second container empty for now (the white box in the image below).

Render this "Home" component from within App.tsx.

https://drive.google.com/file/d/1KyVTR7NxokTEapP8lvRwz2XqT8XUqzFa/view?usp=sharing

PR #55, opened 2020-04-17
Feature/49 create home and sidebar

QuarkNerd:
https://github.com/Furlough-Friends/rota-flex-101/issues/49


frontend/src/components/app/index.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
QuarkNerd, line 1: First thing to note is im using the index convention, which means if you name any file index then importing it from outside is simpler as you just import the containing folder. Please say if you dont like this convention. I like it for the the reason mentioned above and because it means if you hide away if there are containers.

frontend/src/components/sidebar/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import sidebarStyles from './sidebar.module.scss';
+
+const Sidebar = () => {
+  const buttonList = [
+    { name: 'Summary' },
+    { name: 'Rota' },
+    { name: 'Employees' },
```
QuarkNerd, line 8: objects so that later they can be given Ids to tell the app what the user has clicked and it will be nice (i feel) if ids had no spaces whereas menu options might

frontend/src/components/sidebar/sidebar.module.scss
```diff
@@ -0,0 +1,12 @@
+.main {
+    background-color: #106cc8;
```
QuarkNerd, line 2: not sure what the proper way to do this is, but since this file is imported as a module it feels strange to call this class sidebar as then it will be called sidebarStyles.sidebar, so ive opted to say "main". open to other suggestions

JAStrachan:
Hey Guveer, mind writing a fuller description for this pull request! It helps me and Iain to properly review it!

frontend/src/components/home/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+import Sidebar from '../sidebar';
+import homeStyles from './home.module.scss';
+
+const Home = () => (
+  <>
+    <div className={homeStyles.sidebar}>
+      <Sidebar />
+    </div>
+    <div />
+  </>
```
tjanowsk, line 11: Just curious, what's the purpose of empty tags and the empty div>

frontend/src/components/app/index.tsx
JAStrachan: Ah on our last project the Lead thought us to move to the index convention. I think it's an interesting way to do it but it will have to be enforced! I'm happy for this to continue! I'll keep it in mind for future pull requests.

frontend/src/components/sidebar/sidebar.module.scss
tjanowsk, line 2: colours are likely to be reused and since we're using scss it might make sense to pull it as a variable

JAStrachan: Hmmm yeah I get you, I'm not good at names. I've not really used css modules - I've used SCSS instead so this is a little new to me.

frontend/src/components/app/index.tsx
```diff
@@ -0,0 +1,8 @@
...
+import Home from '../home';
+
+const App = () => {
+  return <Home />;
+};
```
tjanowsk, line 6: nitpicky, but this can written more succinctly as
`const App = () => (<Home />)`

frontend/src/components/sidebar/index.tsx
JAStrachan: I am not entirely sure what you mean by this? Also a question could the button styles be determined by the route from the router? That means we have one source of truth.

frontend/src/components/sidebar/sidebar.module.scss
JAStrachan: Actually looking at index.tsx I think naming it sidebar is fine as it means there is no ambiguity.

frontend/src/components/home/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+);
```
floyd924, line 12: When creating React components I would be inclined to write them a little differently. First, delcare the component like this:
`export class Home extends React.Component { }`

Then, inside those curly braces, a React component should always have a `render()` method, like so:
`render() {
        return(
            <div className="home-container">
                <div className="home-left">
                       <Sidebar />
                </div>
               <div className="home-right"></div>
         </div>
        )
    }`

(shame my comment isn't formatted very well, but you get the idea :p )

This also removed the empty `<>`

frontend/src/components/sidebar/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  ];
+  return (
+    <div className={sidebarStyles.main}>
```
floyd924, line 11: If you rewrite the component the same way I've suggested in the `<Home />` component, do you need to use `sidebarStyles.main` or can you give this a className using snake case such as `sidebar-container`?

frontend/src/components/sidebar/index.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import sidebarStyles from './sidebar.module.scss';
+
+const Sidebar = () => {
+  const buttonList = [
+    { name: 'Summary' },
+    { name: 'Rota' },
+    { name: 'Employees' },
+  ];
```
tjanowsk, line 9: It might make sense to pull the hardcoded stuff out of the function and possibly keep it with other constants. That way if we need to change the sidebar we won't have to modify the body of the function which is a small win.

frontend/src/components/sidebar/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <div className={sidebarStyles.main}>
+      {buttonList.map((button) => (
+        <div key={button.name} className={sidebarStyles.menuOption}>
```
floyd924, line 13: Rather than using Divs I think it is better to use a Button element where possible - as @JAStrachan mentioned to me yesterday, as it's better accessibility. Feel free to challenge me on that one though.

After the rewrite of the class, this will fall into the `render()` method. You can either declare 3 Button elements in the html, or you can call a `{this.getButtons}` method in your class from within the render method which will return a list of elements from an array. For example:

Screen Shot 2020-04-17 at 13 26 14

frontend/src/components/sidebar/sidebar.module.scss
floyd924: I think I'd go with something like `sidebar-container` or similar. I'll admint I've not actually seen a className declared using the `className={sidebarStyles.main}` syntax before - was this something you've done on another project?

frontend/src/components/home/index.tsx
QuarkNerd: following the ticket directly which says to create "container" for the the blank part of home

frontend/src/components/sidebar/sidebar.module.scss
QuarkNerd: > Hey Guveer, mind writing a fuller description for this pull request! It helps me and Iain to properly review it! Also what is nice if you are doing front end tickets is to have a screen shot of the relevant change. Firefox has an inbuilt screenshot capture that might make this easier.

is linking the ticket sufficient, in this case I cant think of what else to add.

QuarkNerd: I agree, I didn't do it this time just because I was told (in the ticket) to not worry about colours yet so I figured it can wait, will change it now

frontend/src/components/home/index.tsx
floyd924: I realise it's probs easier with an image. Something like this:

Screen Shot 2020-04-17 at 13 36 38

frontend/src/components/app/index.tsx
QuarkNerd: Good spot, I feel there might be a liniting option for this

frontend/src/components/sidebar/index.tsx
QuarkNerd: so when the buttons are clicked one of two things will happen
1. they each have their own callback (unlikely) so we need an object to call it with
2. they have one given callback function that they call with different values (lets call them IDs), I wanted to separate ID and name because
  a. I feel IDs (entirely internal shoud be camelcase no spaces
  b. separate the display from implementation. 

Im not sure what you mean either haha, but to share button styles I would suggest exporting a @mixin from whereever we define constant colours

frontend/src/components/home/index.tsx
tjanowsk: Is there a reason you would suggest class components as opposed to functional components with hooks?
React seem to recommend the latter https://reactjs.org/docs/hooks-intro.html

frontend/src/components/sidebar/sidebar.module.scss
QuarkNerd: > I think I'd go with something like `sidebar-container` or similar. I'll admint I've not actually seen a className declared using the `className={sidebarStyles.main}` syntax before - was this something you've done on another project?

Yeah, we did this in our grad project, the "main" part is probably not standard as I came up with it and I imagine they just let it slide. The part of having classnames in an object is using modules which is used a lot and very useful as it avoids css collisions without making classnames in your scss file longer than needed. Im happy to not use this if people prefer.

floyd924: I'm happy for you to continue to use it if worked well for you guys on the grad project 👍

frontend/src/components/home/index.tsx
QuarkNerd: the empty tag is a way to stop react from complaining and saying you should only return one react child (it adds nothing to the dom)

frontend/src/components/home/index.tsx
floyd924: Primarily it is familiarity. If you guys would rather use functional and hooks, that's fine by me 👍

QuarkNerd: I've been taught against using class components as  with hooks they no longer provide any extra functionality. All the React mentors at newcastle were against class components and theres been talk of getting functional components in the training doc

floyd924: Yeah that's fair. Proceed

frontend/src/components/sidebar/index.tsx
QuarkNerd: I can do that if you like, all what im doing does is it allows the classnames in the scss file to be like` .container`  or `.button` ect and the fact that it is a module automatically handles adding the sidebar part. Its what I used in my grad project and an accepted concept

frontend/src/components/sidebar/index.tsx
tjanowsk: Slightly off-topic, but can getCardNames be written more simply as (omitting typescript fluff):
`getCardNames = () => cards.map(card => (<p>I just drew a {card.name}</p>))`
or better still
`getCardNames = cards => cards.map(card => (<p>I just drew a {card.name}</p>))`
to make the function pure?

The latter can be written even more nicely using ramda's curried map:
`getCardNames = R.map(card => (<p>I just drew a {card.name}</p>)`

frontend/src/components/sidebar/index.tsx
floyd924: yeah if it worked in grad project it's fine to go ahead

frontend/src/components/home/index.tsx
floyd924: Yeah, react complains if you don't.  rather than an empty tag I would maybe wrap all the contents in a div with a className something like home-container?

frontend/src/components/sidebar/index.tsx
floyd924: yes it can

floyd924:
Last thing I would change is the buttons in the sidebar could be `<button />` elements instead of `<div />`'s as this implies they are clickable and tabbable

frontend/src/components/home/index.tsx
QuarkNerd: For now ive just removed it because it's cleaner and we can add things back in when we need to.

QuarkNerd:
Okay so a couple of questions. Firstly regarding the getcards or as Ive called it getButtons functions, if I define its type of parameter where should that be stored? (right now its just a any[].

Secondly, is there a proper way to bulk remove the button style added by the browser?

JAStrachan:
> 
> 
> Okay so a couple of questions. Firstly regarding the getcards or as Ive called it getButtons functions, if I define its type of parameter where should that be stored? (right now its just a any[].
> 
> Secondly, is there a proper way to bulk remove the button style added by the browser?

So for the any[] thing you should type it with a Typescript type you have defined. Probably in somewhere close to where you use that object. 

For the style are you talking about removing the highlighted style?

JAStrachan:
I think if you have a rerender the css classes get reapplied? If these buttons took you to different routes in the app I suspect that would cause a rerender and might solve the problem. To give other options there is this https://reacttraining.com/reach-ui/tabs which is part of a ui library that can do tabs. Pretty customisible. I think its worth continuing down the path of trying to get the right css to apply though.

floyd924:
>  if I define its type of parameter where should that be stored? (right now its just a any[].

So, it's an array of objects, which have a custom shape. Sounds like we're gonna need to make an interface here. Perhaps inside the `sidebarOptions.js` file. Something like this:

`interface SidebarOption { name: string }`

You should also add this type to the sidebarOptions array that's in that file, like this:

`const sidebarOptions: SidebarOption[] = [`

QuarkNerd:
so for style, I mean the fact that is a button, has style. Do i have to reply properties individually or is there a catch all to that stuff

floyd924:
Looks good to me! If there are specific button styles we want to unset I might use something like `border: unset` in the css file. You can do `all: initial` or `all: unset` to wipe out all inherited styles, too. Are there particular parts which are causing problems just now?

QuarkNerd:
> Looks good to me! If there are specific button styles we want to unset I might use something like `border: unset` in the css file. You can do `all: initial` or `all: unset` to wipe out all inherited styles, too. Are there particular parts which are causing problems just now?

Good to know, thanks. The tickets says not to worry about all the styles just yet so I will leave it out of this PR

Issue #50, opened 2020-04-16
Create initial content components


floyd924:
Requires #49 

Create 3 react components: "Summary" , "Rota" and "Employees". Each should just contain an empty div with a className, and maybe an h1 saying which component it is so we can see which is selected later on.. 

Create an SCSS file for each one, and give each a width and height of 100%.

Add the "Employees" component to the render method of white box in Home.tsx so it renders next to the sidebar, in the white box in this design:

https://drive.google.com/file/d/1KyVTR7NxokTEapP8lvRwz2XqT8XUqzFa/view?usp=sharing

PR #60, opened 2020-04-20
feature 50 - create initial content components

tjanowsk:
Added templates for several components.
Buttons don't change the main screen yet, only the employee screen is visible now.

Tried css grid for the overall layout, this may be changed when collapsible menu is used.

https://github.com/Furlough-Friends/rota-flex-101/issues/50


QuarkNerd:
Your PR seems to be based of mine (49), should remove my commits, using rebase --onto should do the trick. Let me know if you need help

Also rather than having, employees, summary and rota as completely separate components, do you think its worth reusing code? Maybe make a page that is a HOC https://reactjs.org/docs/higher-order-components.html that we call page and we pass in the title and inner display

PR #63, opened 2020-04-21
Add initial components

tjanowsk:
Re-uploading this PR after a failed rebase.

No routing yet, the buttons don't do anything for the moment.

Changed the styling of the main page to use css grid, this may need to be changed with the collapsible  sidebar.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -12,11 +12,13 @@
     "@types/react": "^16.9.32",
     "@types/react-dom": "^16.9.6",
     "@types/react-redux": "^7.1.7",
+    "@types/react-router-dom": "^5.1.4",
     "@types/redux-logger": "^3.0.7",
     "node-sass": "^4.13.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-redux": "^7.2.0",
+    "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
     "redux-logger": "^3.0.6",
     "typescript": "^3.8.3"
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -28,8 +28,8 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "npx prettier --check ./src/**/*.{js,jsx,ts,tsx,scss} && npx eslint ./src/**/*.{js,jsx,ts,tsx}",
-    "format": "npx prettier --write ./src/**/*.{js,jsx,ts,tsx,scss} && npx eslint --fix ./src/**/*.{js,jsx,ts,tsx}"
+    "lint": "npx prettier --check './src/**/*.{js,jsx,ts,tsx,scss}' && npx eslint './src/**/*.{js,jsx,ts,tsx}'",
+    "format": "npx prettier --write './src/**/*.{js,jsx,ts,tsx,scss}' && npx eslint --fix './src/**/*.{js,jsx,ts,tsx}'"
   },
   "eslintConfig": {
     "extends": "react-app"
--- a/frontend/src/components/employees/employees.module.scss
+++ b/frontend/src/components/employees/employees.module.scss
@@ -1,0 +1,7 @@
+.employees {
+    background-color: olive;
+    color: orange;
+    grid-area: main;
+    width: 100%;
+    height: 100%;
+}
--- a/frontend/src/components/employees/employees.module.scss
+++ b/frontend/src/components/employees/employees.module.scss
@@ -1,7 +1,7 @@
 .employees {
-    background-color: olive;
-    color: orange;
-    grid-area: main;
-    width: 100%;
-    height: 100%;
+  background-color: olive;
+  color: orange;
+  grid-area: main;
+  width: 100%;
+  height: 100%;
 }
--- a/frontend/src/components/employees/index.tsx
+++ b/frontend/src/components/employees/index.tsx
@@ -1,0 +1,11 @@
+import React from 'react';
+import employeesStyle from './employees.module.scss';
+
+const Employees = () => (
+  <div className={employeesStyle.employees}>
+    <h1> Employees </h1>
+    <p> Employees component </p>
+  </div>
+)
+
+export default Employees;
--- a/frontend/src/components/employees/index.tsx
+++ b/frontend/src/components/employees/index.tsx
@@ -6,6 +6,6 @@ const Employees = () => (
     <h1> Employees </h1>
     <p> Employees component </p>
   </div>
-)
+);
 
 export default Employees;
--- a/frontend/src/components/home/index.tsx
+++ b/frontend/src/components/home/index.tsx
@@ -1,9 +1,11 @@
 import React from 'react';
 import Sidebar from '../sidebar';
+import Employees from '../employees';
 import homeStyles from './home.module.scss';
 
 const Home = () => (
-  <div className={homeStyles.sidebar}>
+  <div className={homeStyles.home}>
+    <Employees />
     <Sidebar selctedOption="Summary" />
   </div>
 );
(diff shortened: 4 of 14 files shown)


frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,11 @@
...
+import employeesStyle from './employees.module.scss';
+
+const Employees = () => (
+  <div className={employeesStyle.employees}>
+    <h1> Employees </h1>
+    <p> Employees component </p>
+  </div>
+);
+
+export default Employees;
```
QuarkNerd, line 11: Given the repetition of code is it worth using a HOC for employee, Rota, and summary?

frontend/src/components/employees/employees.module.scss
```diff
@@ -0,0 +1,7 @@
+.employees {
+    background-color: olive;
+    color: orange;
+    grid-area: main;
+    width: 100%;
+    height: 100%;
+}
```
QuarkNerd, line 7: if we do not switch to HOCs or some way to reduce code repetition, maybe we should use a mixin to avoid repeating lines 4-6?

frontend/src/components/employees/index.tsx
tjanowsk: There will be no repetition once the components are fully implemented (I think?).
These are just placeholders.

frontend/src/components/employees/employees.module.scss
tjanowsk: As above, these are going to change once the components are properly implemented. 
If there is repetition in the real thing we can always refactor later.

frontend/src/components/employees/index.tsx
JAStrachan: I agree for the now yeah they are just placeholders. I think its worth waiting before too early refactoring.

JAStrachan: Though I think we might get there one day due to styling for the collapsible sidebar

JAStrachan:
CSS grid will work for the collapsible sidebar don't worry, its just changing grid columns!

frontend/src/components/employees/employees.module.scss
QuarkNerd: It feels very likely that the repetition will remain. Especially if we have a title/header. If people feel strongly I will remove my changes requested but it will be a lot easier to reaactor now. Do we have designs for what the components will contain?

frontend/src/components/employees/index.tsx
QuarkNerd: do we have any designs for the different pages yet?

frontend/package.json
```diff
@@ -26,8 +28,8 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "lint": "npx prettier --check ./src/**/*.{js,jsx,ts,tsx,scss} && npx eslint ./src/**/*.{js,jsx,ts,tsx}",
-    "format": "npx prettier --write ./src/**/*.{js,jsx,ts,tsx,scss} && npx eslint --fix ./src/**/*.{js,jsx,ts,tsx}"
+    "lint": "npx prettier --check './src/**/*.{js,jsx,ts,tsx,scss}' && npx eslint './src/**/*.{js,jsx,ts,tsx}'",
+    "format": "npx prettier --write './src/**/*.{js,jsx,ts,tsx,scss}' && npx eslint --fix './src/**/*.{js,jsx,ts,tsx}'"
   },
```
QuarkNerd, line 33: what does this change do?

frontend/src/components/home/index.tsx
```diff
@@ -1,11 +1,23 @@
...
-  <div className={homeStyles.sidebar}>
-    <Sidebar />
-  </div>
+  <Router>
+    <div className={homeStyles.home}>
+      <Route path="/employees" component={Employees} />
+      <Route path="/rota" component={Rota} />
+      <Route path="/summary" component={Summary} />
+      <Route path="/" component={Sidebar} />
+    </div>
```
QuarkNerd, line 16: Is there a reason we have the sidebar inside the router? I know nothing about best practices but what if we had the sidebar outside the router (given its always there so the two dont need to talk)

tjanowsk: Sidebar contains <Link> elements that redirect to other pages and must be inside router.
Otherwise I think I'd have to do something ugly like write some javascipt to change the url manually.

frontend/package.json
tjanowsk: My shell was trying to expand the paths before passing them to npx resulting in an error.
Adding quotation marks means the shell treats it as a string and the command runs correctly.

frontend/src/components/home/index.tsx
emab: Could the router in the `App` component instead of `Home`? That way all components in Home would be inside Router, or would this not work?

floyd924:
> Looks good to me, for initial components this seems to cover all of the bases.
> 
> The only thing I would say is that, once we start using a framework for component styling, we shouldn't need much CSS. In my (limited) experience I've been told to only use CSS if you cannot get something to display correctly - otherwise the framework should be used.
> 
> I appreciate the CSS is here now to give the page a bit of structure.

By this do you mean specifically the layout, so we don't have to specify how containers respond, or the specific styling such as colours/margins?

I guess we don't really have UX here telling us specific colours or pixel widths to use, so working with a framework is maybe our best bet,  though I've not used many myself. 

Perhaps an alternative would be to create several base classes, or mixins, which would style all the buttons, or divs, a certain way across our app, applying specific css properties we want to evey instance.

emab:
> > Looks good to me, for initial components this seems to cover all of the bases.
> > The only thing I would say is that, once we start using a framework for component styling, we shouldn't need much CSS. In my (limited) experience I've been told to only use CSS if you cannot get something to display correctly - otherwise the framework should be used.
> > I appreciate the CSS is here now to give the page a bit of structure.
> 
> By this do you mean specifically the layout, so we don't have to specify how containers respond, or the specific styling such as colours/margins?
> 
> I guess we don't really have UX here telling us specific colours or pixel widths to use, so working with a framework is maybe our best bet, though I've not used many myself.
> 
> Perhaps an alternative would be to create several base classes, or mixins, which would style all the buttons, or divs, a certain way across our app, applying specific css properties we want to evey instance.

I think a bit of both, and also by using a framework (or creating our own styled components) it allows better levels of abstraction in a component. Basically it allows a component to make this transformation ([Image credit](https://medium.com/the-non-traditional-developer/styling-best-practices-using-react-c37b96b8be9c)):

image

frontend/src/components/sidebar/sidebar.module.scss
```diff
@@ -2,6 +2,7 @@
 
 .container {
   background-color: $colour-darkBlue;
+  grid-area: sidebar;
```
QuarkNerd, line 5: Could you explain the reasoning to use grid area? Rather than wrapping the sidebar in a div with a width from the outside. It feels like doing it this way makes, styling the component implementation dependent? In that you have to know internal details to use it

tjanowsk: That's how grid is used in general, the parent components says where each child component should be placed on the grid and child components are identified by name.

I don't think of css as an implementation detail. Any user can preview the element's style in the browser and any css can always be overridden.
The only question is where to put the child component's grid name css, which could be either with the component or with home.

The upside of using grid is that you don't need any extra divs just to get the layout right, which helps to keep the html clean.

floyd924: Will it be straightforward to hide the sidebar on a narrow window with grid?

tjanowsk: I haven't tried, but you should be able to by wrapping it inside `@media` block like shown
https://webmadewell.com/css-grid-responsive-layouts/

Issue #52, opened 2020-04-16
Add styling to sidebar


floyd924:
Requires #49

Style the sidebar as per designs.

Dark blue colour is: #106cc8
When the component is selected/rendered it should be light blue, code: #3e7dbd
When the button is hovered over, the cursor should change. Perhaps use `&:hover{ cursor: pointer }`

There is so specific font, button height, or line height specified, so this is somewhat up to you. Make it similar-ish to the design, and make it look nice!

Designs:
https://drive.google.com/drive/folders/1Sb2xj92dNSkCcfJKjQqj1d-aay3whTNx?usp=sharing

PR #64, opened 2020-04-21
Feature/52 add colours to sidebar

QuarkNerd:
https://github.com/Furlough-Friends/rota-flex-101/issues/52

I need guidance on the font. 

image


frontend/src/components/sidebar/index.tsx
```diff
@@ -2,15 +2,22 @@ import React from 'react';
...
 import sidebarOptions, { SidebarOption } from '../../constants/sidebarOptions';
 
-const getButtons = (options: SidebarOption[]) =>
+const getButtons = (options: SidebarOption[], selctedOption: String) =>
   options.map((o) => (
-    <button type="button" className={sidebarStyles.menuOption} key={o.name}>
+    <button
+      type="button"
+      className={`${sidebarStyles.menuOption} ${
+        o.name === selctedOption ? sidebarStyles.selectedMenuOption : ''
```
floyd924, line 10: If the second part of a statement like this is null, or an empty string, you could also write it like `${o.name === selectedOption && sidebarStyles.selectedMenuOption}`

frontend/src/components/sidebar/index.tsx
```diff
@@ -2,15 +2,22 @@ import React from 'react';
...
+      }`}
+      key={o.name}>
       {o.name}
     </button>
   ));
 
-const Sidebar = () => (
-  <div className={sidebarStyles.container}>{getButtons(sidebarOptions)}</div>
+const Sidebar = ({ selctedOption }: { selctedOption: string }) => (
```
floyd924, line 17: Are the curly braces neccessary around the `selectedOption` and `selectedOption: string`?

frontend/src/components/sidebar/index.tsx
QuarkNerd: The html changes to this ```<button type="button" class="sidebar_menuOption__12WdO false">Rota</button>```   We have a false in the class. I think the reason this doesnt work is what you are suggesting is a JSX thing but doesnta pply to JS logic in strings?

frontend/src/components/sidebar/index.tsx
QuarkNerd: I believe so, because the function is called with a "props object" which I am deconstructing

frontend/src/components/sidebar/index.tsx
floyd924: Yeah I think you're right. That's annoying

Issue #41, opened 2020-04-09
Authentication setup


mangan360:
As a PO, I want infrastructure in place so users can sign into the application through an authorization system. Only **active users** will be able to login.

Acceptance Criteria: 
-   Given when a user is an employee of the company, when they attempt to login to the application then they are brought to the landing page of the app. 
- Given when a user who isn't an employee of the company, when they attempt to login to the application then they are prompted with a message "You don't have access to this application"

duncstin:
There are a number of elements involved with this task, allowing it to be broken down into a series of logical steps. Writing this with the assumption we're using Auth0, but if we choose a different service the same approach should be adopted: [Authorization Code Flow with Proof Key for Code Exchange](https://auth0.com/docs/flows/concepts/auth-code-pkce). (Worth noting: some guides may suggest using the [Implicit Flow](https://auth0.com/docs/flows/concepts/implicit), but this is only appropriate for logging in, and is not longer considered best practice to obtain access tokens).

A rough breakdown of tasks involved is as follows.

- Signup for Auth0, and [configure it appropriately](https://auth0.com/docs/architecture-scenarios/spa-api/part-2). This involves:
  - signing up to Auth0,
  - Adding the API, and configuring scopes,
  - Adding the frontend (SPA) and configuring permissions
  - (Optionally, create 2 dummy users, an Admin and a normal employee)
You may need to read up on Auth0's [scopes](https://auth0.com/docs/scopes/current/api-scopes#example-an-api-called-by-a-first-party-application) and [roles](https://auth0.com/docs/authorization/concepts/rbac).
Also note that setup for front and back end is shown a bit in links below.

- Authenticate the user on the front end. This boils down to obtaining an **Access Token** for the current user, which will be needed to access our API. This task is simplified by the existence of [Auth0's SPA SDK](https://auth0.com/docs/libraries/auth0-spa-js). For example usage with react, see the [quickstart](https://auth0.com/docs/quickstart/spa/react), which introduces some custom hooks. 

- Secure the backend. The backend should check for a valid access token for all endpoints. Additionally, it should ensure only an Admin can access admin specific endpoints. See the [example using Spring](https://auth0.com/docs/quickstart/backend/java-spring-security5). Note that when [validating the access token](https://auth0.com/docs/tokens/guides/validate-access-tokens).

[...]

PR #65, opened 2020-04-21
Feature/41 authentication setup

emab:
### [#41 Authentication Setup](https://github.com/Furlough-Friends/rota-flex-101/issues/41)
I've worked on the first bullet point mentioned by Duncan.

A user can log in via the button. It redirects to Auth0, and once successfully logged in the user is returned to the app. The logged in user now has a valid session and token.

The `react-auth0-spa.tsx` file is provided by Auth0, and creates the Auth0 client and gives access to various hooks. These are accessed by `import { useAuth0 } from './react-auth0-spa'.

I've not looked at different roles yet, but I'm assuming by assigning roles through the Auth0 API we can work out what role a user has via their token.

____________________________________________

Apologies for the amount of commits made in order to pass the GitHub checks. It's my first time dealing with those, and next time I'll do more checks locally before creating a PR!


frontend/src/components/loginOutButton/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const {
+    loading,
+    isAuthenticated,
+    loginWithRedirect,
+    logout,
+    user,
+  } = useAuth0();
+
+  return (
+    <>
```
floyd924, line 17: I'd maybe make this a div, with a descriptive className, so that it's easier to identify in the future

frontend/src/index.tsx
```diff
@@ -1,14 +1,32 @@
...
 import { Provider } from 'react-redux';
+import { Auth0Provider } from './react-auth0-spa';
 import App from './components/app';
+import config from './auth_config.json';
+import history from './utils/browserHistory';
 import { store } from './app/store';
 import * as serviceWorker from './serviceWorker';
 
+// Routes user to correct url after authenticating
+const onRedirectCallback = (appState: any) => {
```
floyd924, line 12: In Typescript it's a good idea to avoid `any` type as much as possible. My guess would be we'll need to make an interface for appState, but does it work if we take `appState: any` out completely?

emab: Okay good to know - originally it had no type there but that didn't work. I'll create an interface and get that sorted.

emab: I've updated this to use an interface. It works (I've tested redirecting somewhere and it lands you at the page you want).

AdrianBorg:
not had a chance to look through your whole PR yet, but something about this:

> Apologies for the amount of commits made in order to pass the GitHub checks. It's my first time > dealing with those, and next time I'll do more checks locally before creating a PR!

1) Don't worry about number of commits, it'll force you to squash and merge here on the PR before you merge to master, so it'll be completely fine

2) If you want the checks to run automatically before any local commit, we've set up git pre-commit hooks that are easy to configure, there should be the instructions of 2 ways to do on the readme in the root directory. It just requires 1-2 lines to point git to the right location of the hooks

frontend/src/components/loginOutButton/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+          onClick={() => loginWithRedirect({})}>
+          Log in
+        </button>
+      ) : (
+        <button
+          type="button"
+          disabled={loading}
+          onClick={() => logout({ federated: true })}>
+          Log out {!loading && user.name}
+        </button>
```
floyd924, line 31: There is a little code duplication here. It might be neater to just render 1 button, with the `isAuthenticated ?` check either happening in the onClick method and the button text, or extracted into it's own methods for each. Not sure which you'd prefer or which you think would look better

frontend/src/react-auth0-spa.tsx
```diff
@@ -0,0 +1,124 @@
...
+    initAuth0();
+    // eslint-disable-next-line
+  }, []);
+
+  const loginWithPopup = async (o: PopupLoginOptions) => {
+    setPopupOpen(true);
+    try {
+      await auth0Client!.loginWithPopup(o);
+    } catch (error) {
+      console.error(error);
```
floyd924, line 84: Not sure what people's opinions are on the Github Actions warning being thrown here @AdrianBorg @tjanowsk ?

emab: Yeah I was going to ask about this. The only reason I left it here is because it was how it was shipped with Auth0.

frontend/src/components/loginOutButton/index.tsx
emab: I've had a look and changed this.

frontend/src/react-auth0-spa.tsx
tjanowsk: According to Laurie leaving console statements is a no-no.
We probably want to give meaningful feedback to the user anyway and most of them don't look at the console.
One possibility is to use `toastr` to display the error message, I used it for the grad training project and it's quite nice.
It's not necessarily part of this PR though.

frontend/src/auth_config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "domain": "rota-flex-101.eu.auth0.com",
+    "clientId": "NqmZExdF9Q14mgi7p5o5RkPt338B2irO"
```
tjanowsk, line 3: I don't know much about auth0, so just wanted to make sure that it's ok to commit clientId to the repo, i.e. it's not sensitive information?

frontend/src/utils/browserHistory.ts
```diff
@@ -0,0 +1,3 @@
+import { createBrowserHistory } from 'history';
+
+export default createBrowserHistory();
```
AdrianBorg, line 3: Whats the purpose of this file? Is it just to rename the import in `index.tsx`?

frontend/src/index.tsx
```diff
@@ -1,14 +1,38 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
+import { RedirectLoginResult } from '@auth0/auth0-spa-js';
+import { Auth0Provider } from './react-auth0-spa';
 import App from './components/app';
+import config from './auth_config.json';
+import history from './utils/browserHistory';
```
AdrianBorg, line 8: would `import { createBrowserHistory as history } from 'history'` work the same?

frontend/src/auth_config.json
emab: I'm pretty sure the clientId can be here, as in Auth0 you define where requests are allowed to come from. So only requests from our app can authenticate with Auth0. There is a secret client ID which should definitely not be shared. I think this public clientId could allow third party apps to use our API for example.

frontend/src/utils/browserHistory.ts
tjanowsk: createBrowserHistory is a function, here you're calling it and returning the output - whatever it is
Doesn't look like renaming, more like initialising?

emab: Yeah this just initialises it, and a few resources online seem to do this by creating it in the utils folder. I'm going to rename it back to `history` as this also seems to be common practice.

This is also how Auth0 use it in their documentation when setting up the front-end.

frontend/src/index.tsx
emab: I guess so, it doesn't look wrong. I think it's just common practice to create it in the utils and use it from there.

Pretty much every website that talks about using the history module seems to have a class which imports that method and then exports default. I think it's something to do with history being a global API and should be accessible by all of the app, and by importing it there, all components are using the same instance of the method.

tjanowsk: It's not the same for the same reason as the other comment:
in `import { createBrowserHistory as history } from 'history'`, history is a function,
in `import history from './utils/browserHistory'`, history is the object that this function returns.

emab: Yep that sounds right. Either way - it's not something I came up with, it's what the Auth0 and other docs (such as react-router?) do when using the history package.

frontend/src/react-auth0-spa.tsx
floyd924: I reckon we can leave this in for now. A console error is better than a console log, which implies the code hasn't been tidied up

emab:
Sorry - had to fix conflicts before merge.

Issue #33, opened 2020-04-09
Create getMyShifts endpoint


mangan360:
As an employee I want to be able to get all my shifts for a given week. By using an API tool such as Postman, when I hit the getEmployeeShifts endpoint with a specified date range (a week), then I should get back my shifts for that week.

If we do not yet have token authentication in place, this endpoint will need to take an employee's ID in the request header, or as a query param. Later this will be extracted from the tokem. 

Acceptance Criteria : 
- Given a employee with basic access. When they make a request for their shifts between specified dates then they should receive shift back. Each shift should contain : date, start & end time and the shift type (an employee may need to work in a different area).

Issue #30, opened 2020-04-09
Create user endpoint


mangan360:
As a manager, when a new employee starts a job I want to be able to add them to the existing database. There should be an endpoint in place which takes user information as a payload and saves. A unique ID should be generated for this user and they should then saved in the database. 

Acceptance Criteria: 
- Given a user with admin (manager) access, when the manager makes a request to the createUser endpoint with the correct payload then a new user should be created in the database with a unique ID. The manager should receive an OK response.  
- Given a user with employee access, when the employee makes a request to the createUser endpoint, they should be get a bad response with a message stating that they don't have the correct access

floyd924:
An employee trying to access this endpoint without Manager status shoudl receive 403 Forbidden

PR #66, opened 2020-04-21
Feature/30 create user endpoint

AdrianBorg:
#30 Create User Endpoint

Created the `.../staff/create` POST endpoint

Before this commit there was no way of adding a staff entity into the db. This ticket aimed at adding this functionality, allowing a Manager to create a staff but rejecting the request if a regular user does it.

I've based this off of #54 (the PR for #33) so it will need to be merged to master after that one is merged. It uses the same class to mock out the authentication for now and will need to be replaced once we have Auth0 set up for the backend

I've created a `CreateStaffForm` to act as a DTO object anticipating that the input would be coming from a manager filling out a form on the front end. This makes use of java validation to ensure the firstname, surname and role are not absent from the POST request body.

Also changed the `Staff` obejct field `active` to `inactive` this didn't matter much to anything else, but makes the handling of making it default to creating an active user much easier (it'll be automatic).


backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -12,17 +13,37 @@
...
+   * TODO Temp solution needs to be changed to be more appropriate when auth is added to backend.
+   * Get user role from token.
+   * @param token Authentication token
+   * @return {@link Role} of the user or <code>null</code> if token is not valid.
+   */
+  public static Optional<Role> getUserRoleFromToken(String token) {
     return validateToken(token)
-        ? Optional.of(token)
+        ? Optional.of(
+          (token.length() % 2 == 0)
```
floyd924, line 43: incase we add additional roles later on, would you be better checking if the list contains Role.Manager?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -12,17 +13,37 @@
    * @return Whether authentication was successful.
    */
   public static boolean validateToken(String token) {
-    return true;
+    return !token.toLowerCase().equals("false");
```
floyd924, line 16: Can you explain what's going on on this line?

backend/src/main/resources/data.sql
```diff
@@ -1,10 +1,10 @@
...
-active)
+inactive)
 VALUES
-    ('Joe', 'Bloggs', '2018-08-03', 40, 1.23, 'USER', 'barista', true),
-    ('Mary', 'McGee', '2017-08-12', 40, 999.99, 'MANAGER', 'supervisor', true),
-    ('Json', 'Server', '2015-01-12', 40, 0.01, 'USER', 'floor', true),
-    ('Arthur', 'Guinness', '2019-12-12', 40, 2.99, 'USER', 'chef', true);
+    ('Joe', 'Bloggs', '2018-08-03', 40, 1.23, 'USER', 'barista', false),
+    ('Mary', 'McGee', '2017-08-12', 40, 999.99, 'MANAGER', 'supervisor', false),
+    ('Json', 'Server', '2015-01-12', 40, 0.01, 'USER', 'floor', false),
```
floyd924, line 6: surely this has to be Jason? 😂

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -0,0 +1,51 @@
...
+   * @param start start date or null to get from the beginning
+   * @param end end date
+   * @return A list of member's engagements
+   */
+  public List<EngagementDto> getStaffEngagementsBetween(int staffId, Instant start, Instant end) {
+    Instant startTime = start == null ? Instant.MIN : start;
+    Instant endTime = end == null ? Instant.MAX : end;
+    return engagementRepository.findByStaffId(staffId).stream()
+        .filter(engagement ->
+            engagement.getStart().isAfter(startTime)
```
floyd924, line 36: Should we make this 'after or equal to startTime', etc?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
AdrianBorg: This makes it so that we have a way of providing an invalid token. 
If we pass in `false` as the token then it'll output false. Otherwise it'll be true.
(not sensitive to case)

Maybe i should add a comment there, or split it up in to separate lines to make it clearer

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
AdrianBorg: Wasn't sure how we're going to manage roles, and didn't want to just take a decision here in case it makes it harder later. Need to decide if it'll depend on using Auth0 roles (and setting them up) or based off of the roles as in our db. Also we might have to somehow synchronise the 2. 

This was just a placeholder and should be changed once we decide that.

backend/src/main/resources/data.sql
AdrianBorg: I'm guessing this is another one of @tjanowsk 's easter eggs. I'm loving them! Can't believe i missed it in his PR.
I'm sure it was specifically for @Skoomaki

backend/src/main/java/com/rota/api/StaffService.java
AdrianBorg: I've not actually made these changes in my PR. But I think it'll be a good spot to change them over since that PR is already merged. I'll also change the Before one too.

@tjanowsk would you agree? or is it purposefully as only after/before.

backend/src/main/resources/data.sql
tjanowsk: As much as I'd like to take credit, this is actually taken verbatim from
https://docs.google.com/document/d/1amxWzf-OZsJokOBTwgHSoibYUACiPXb0IEyx-cOv0SY/edit
so it's @smangan360 easter egg

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: I don't mind, for continuous variables like time there's not much difference between greater and greater or equal.

backend/src/main/java/com/rota/api/dto/form/CreateStaffForm.java
```diff
@@ -0,0 +1,65 @@
+package com.rota.api.dto.form;
```
tjanowsk, line 1: Why do we need both this and StaffDto class?

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -28,10 +33,21 @@
     Instant endTime = end == null ? Instant.MAX : end;
     return engagementRepository.findByStaffId(staffId).stream()
         .filter(engagement ->
-            engagement.getStart().isAfter(startTime)
-          && engagement.getEnd().isBefore(endTime))
+            (engagement.getStart().equals(start)
+                || engagement.getStart().isAfter(startTime))
```
tjanowsk, line 37: This can be refactored to
`!engagement.getStart().isBefore(startTime)`
that way you only call the comparison once. Not sure about the readability though.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -56,4 +64,39 @@
...
+      @ApiParam("Staff details for the new user")
+          CreateStaffForm createStaffForm
+  ) {
+    final Role role = AuthenticationUtils
+        .getUserRoleFromToken(authString)
+        .orElseThrow(() ->
+          new ResponseStatusException(HttpStatus.UNAUTHORIZED, "Failed to authorize.")
+        );
+
+    if (role != Role.MANAGER) {
```
tjanowsk, line 94: I would replace the content of this if statement with a function e.g. `hasManagerPermissions(role)`.
That way if we ever add more roles that have manager permissions we'll be able to fix that easily.

backend/src/main/java/com/rota/api/StaffService.java
floyd924: If we say 'show me everything from Monday 9am' we want all shifts that start at 9am on monday so I'd be inclusive here

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
floyd924: Sure. Let's discuss roles at the call today if enough people are online

emab: We can give/get users roles via the [Auth0 Management API](https://auth0.com/docs/api/management/v2#!/Users/get_user_roles). 

The access token used to make requests from the front-end will contain the users roles (scope) in its payload.

PR #67, opened 2020-04-22
Feature/41b authentication setup backend

emab:
### [#41 Authentication Setup Backend](https://github.com/Furlough-Friends/rota-flex-101/issues/41)
This PR continues on from the [previous PR](https://github.com/Furlough-Friends/rota-flex-101/pull/65). It adds support in the back end to verify a Jwt token sent by the front end.

Endpoints can be secured in `SecurityConfig`. Currently all API endpoints (apart from swagger) require authorisation, and as well as authorisation via access token a user must also have sufficient role in the back-end database to access endpoints. Endpoints that require this kind of authorisation can also be configured in `SecurityConfig`.

#### Summary
The back end no longer uses the Auth0 Management API. This means the client secret is no longer needed. Instead we use the access token and Auth0 `userInfo` to control access.

#### Changes
We decided that user roles will be stored in the back-end database. Now, when a user sends a request to the back-end API, their email address is extracted using the Auth0 ID in the access token and Auth0's `userInfo` API endpoint. A user is **authenticated if they log in via Auth0** but will not have access **unless they are in the back-end database with a specific `Role`**. The user role is checked once the user has been authorised (valid JWT token).

We should probably add a screen in the front-end for users who register but do not have their email address in the back-end database. Potentially we could have an invite only system where managers add a user, and this sends an email to the user inviting them to activate their account with Auth0 and linking themselves to the back-end. 

## Update 05/05/2020
I've refactored a lot of the Authentication code to use Spring Security instead of manually passing the token around. We can still access the token value using a method in the `Authentication` class using `getTokenValue()`.

We still need to send some requests to the Auth0 API in order to get `userInfo` which contains the users email. We can use the email (or the users Auth0 ID) to get the users Roles from the Auth0 database.

Since we are using the users email to find them in the database, we will always need use the Auth0 ID from the access token to send a request to Auth0 for the users email. We can tailor the request to the Auth0 API in order to make the response smaller.

I've currently got very vague descriptions for some of the thrown exceptions in the class. The exceptions are thrown by the Java `HttpClient.send()` method.

## Back-end
In order for the back-end to access certain parts of the Auth0 Management API, a `client_secret` must be used to get an `access_token`.

To do this, see rename the file `resources/application-local-example.properties` to `resources/application-local.properties` and add the client secret here. This can be found in [the Auth0 dashboard](https://manage.auth0.com/dashboard/eu/rota-flex-101/applications/CDoLKdSTUvqv50LxjTxvRaHl742WQsyb/settings).

You then need to add `-Dspring.profiles.active=local` to your VM options in the run configuration:
image

This token currently gives us full scope access which we should reduce in the future. Using this access token, we can access all of the [Auth0 Managament API](https://auth0.com/docs/api/management/v2).

### Other points for back-end
Currently, a lot of the methods in `AuthenticationUtils` take a `String token` parameter. We then perform checks on this string. However, when we use Spring Security to protect an endpoint in `SecurityConfig` it automatically stores information about the access token. This can then be accessed using `SecurityContextHolder.getContext()` in the **current execution thread**. This means that the users token doesn't need to be specified in the method, and we don't need to find the "Authorized" header from the request. However, if we are outside the execution thread of the endpoint, then it cannot be accessed. 

### TODO: Back-end
Split `AuthenticationUtils` into two separate classes, one to handle some of the business logic and another to handle any Auth0 related requests. The methods here also need updating to return the correct values, currently I'm not sure what should be returning what.

## Front-end
From the front-end, it is possible to fetch the `/api/public` endpoint, but in order to access the `/api/private` end-point, this header must be used:

`Authorization: Bearer <TOKEN>`

Where token is the logged in user token. This can be obtained via Auth0 hooks and used like this:

```
import React from 'react';
import { useAuth0 } from '../react-auth0-spa';

const ExampleComponent = () => {
  const { getTokenSilently } = useAuth0();

  const getPrivateEndpoint = async () => {
    const token = await getTokenSilently();
    const response = await fetch('http://localhost:3000/api/private', {
      headers: { Authorization: `Bearer ${token}` },
    });
    return response.json();
  };

  return <div>{getPrivateEndpoint()}</div>
};

export default ExampleComponent;
```

I've also temporarily added a button which prints the current users access token to the console to be used in postman API tests for example.


floyd924:
Looks like you're on the right track with this. If a token is found to be invalid, we need to be able to return an informative warning about why it's been rejected. eg, has it expired, was it issued by our issuer for this app, does it have all the parts required for decoding and user pairing, etc

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,219 @@
...
+    HttpRequest request = HttpRequest.newBuilder()
+        .GET()
+        .uri(URI.create("https://rota-flex-101.eu.auth0.com/userinfo"))
+        .setHeader("Authorization", "Bearer " + token)
+        .build();
+
+    HttpResponse<String> response;
+    try {
+      response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+    } catch (IOException | InterruptedException e) {
```
tjanowsk, line 59: This piece of code is repeated a few times so it should be refactored as a separate function.
We may want to convert AuthenticationUtils to Authentication class and keep the response as a class variable. That way we won't have to send requests to auth0 over and over again.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,219 @@
...
     return validateToken(token)
         ? Optional.of(
-          (token.length() % 2 == 0)
+        (token.length() % 2 == 0)
             ? Role.MANAGER
             : Role.USER
-          )
+    )
         : Optional.empty();
   }
```
tjanowsk, line 109: Just marking this as a TODO. We want to extract the role from the token.

backend/src/main/java/com/rota/auth/AudienceValidator.java
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  public OAuth2TokenValidatorResult validate(Jwt token) {
+    OAuth2Error error = new OAuth2Error("invalid_token", "The required audience is missing", null);
+
+    if (token.getAudience().contains(audience)) {
+      return OAuth2TokenValidatorResult.success();
+    }
+    return OAuth2TokenValidatorResult.failure(error);
+  }
```
tjanowsk, line 22: Not sure if it would make this easier to read, but we can also write this as
```
return token.getAudience().contains(audience)
  ? OAuth2TokenValidatorResult.success()
  : OAuth2TokenValidatorResult.failure(
      new OAuth2Error("invalid_token", "The required audience is missing", null));
```

backend/src/main/java/com/rota/api/ExamplePrivateApi.java
```diff
@@ -0,0 +1,22 @@
+package com.rota.api;
+
+import org.springframework.web.bind.annotation.CrossOrigin;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(path = "api")
+@CrossOrigin(origins = "*", allowedHeaders = "*", allowCredentials = "true")
```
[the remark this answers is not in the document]
emab: Is that PR in master? I'll rebase onto it before merging if not.

backend/src/main/java/com/rota/api/ExamplePrivateApi.java
```diff
@@ -0,0 +1,22 @@
+package com.rota.api;
```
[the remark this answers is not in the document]
emab: No - I was just using this class to show how `SecurityConfig` worked with controllers.

.gitignore
```diff
@@ -8,3 +8,4 @@ build
 *.iml
 .vscode
 node_modules
+backend/src/main/resources/application-local.properties
```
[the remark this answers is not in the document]
emab: Yep. I've put some instructions on how to obtain the `client_secret` in my PR description.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
emab: Yeah good point, but by sending them again we validate that the token hasn't expired and is valid. Should definitely be looked into though.

backend/src/main/java/com/rota/auth/AudienceValidator.java
emab: If people find that easier to read I'm happy to make this refactor.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
+
+  /**
+   * Uses the provided access token to get user information from Auth0.
+   *
+   * @param token Valid authentication token.
+   * @return JsonNode of Auth0 information
+   */
+  public JsonNode getUserJsonFromToken(String token) {
```
tjanowsk, line 106: can this be made private?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
+        .setHeader("Authorization", "Bearer " + token)
+        .build();
+
+    HttpResponse<String> response = sendRequest(request);
+
+    try {
+      return objectMapper.readTree(response.body());
+    } catch (JsonProcessingException e) {
+      throw new JsonParseException(e);
+    }
```
tjanowsk, line 119: Instead of catching an exception and throwing a runtime exception, can we omit the try-catch block and have the exception propagate down the stack, i.e.
`public JsonNode getUserJsonFromToken(String token) throws JsonProcessingException`?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
+    return sendRequest(request).body();
+  }
+
+  /**
+   * Get the Auth0 Management API access token - this is needed to get users information, set and
+   * get roles. For this to work, the client_secret must be available.
+   *
+   * @return Auth0 Management API access token.
+   */
+  public String getAuth0ManagementToken() {
```
tjanowsk, line 171: `private`?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
+   *
+   * @return Whether the current token is valid.
    */
-  public static boolean validateToken(String token) {
-    // This will only return false if the token is "false" (case insensitive)
-    return !token.toLowerCase().equals("false");
+  public boolean validateToken(String token) {
+    HttpRequest request = HttpRequest.newBuilder()
+        .GET()
+        .uri(URI.create("https://rota-flex-101.eu.auth0.com/userinfo"))
```
tjanowsk, line 57: "https://rota-flex-101.eu.auth0.com/" is repeated in a few places, I suggest pulling it out as a constant.

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -61,4 +61,19 @@ public Staff createStaff(Staff newStaff) {
...
+   * @param email the user email
+   * @return the {@link Staff} member
+   */
+  public Staff findStaffByEmail(String email) {
+    return staffRepository.findAll().stream()
+        .filter(staff -> staff
+            .getEmail()
+            .equalsIgnoreCase(email))
+        .findFirst()
+        .orElse(null);
```
tjanowsk, line 77: Instead of having orElse(null), I would advise to have the method return Optional<Staff> and use `Optional.ofNullable`. This would immediately indicate to whoever wants to use this method that getting staff by email may fail.

backend/src/main/java/com/rota/api/ExamplePrivateApi.java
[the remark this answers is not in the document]
tjanowsk: No, it's in the other active PR.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
tjanowsk: Can we extract the expiration time and check it ourselves?
I think limiting the traffic with auth0 would be good for our performance (and maybe costs once we're on google cloud).

emab: I've just been testing expiration stuff. The built in `SecurityContextHandler` seems to manage this itself and the token is invalid if expired. I tested this by setting the expiration time as 60 seconds and making repeated calls to the API with the same token.

The token stopped working after the expiration time (although it wasn't exact, was sometimes a few seconds late?).

This means that we can rely on `SecurityContextHolder` to only accept valid, non expired tokens for the current execution.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
-    Integer parsedToken;
-    try {
-      parsedToken = Integer.parseInt(token);
-    } catch (NumberFormatException e) {
-      return Optional.empty();
-    }
-    return Optional.of(parsedToken);
+  public Staff getUserFromToken(String token) {
+    JsonNode userInfo = getUserJsonFromToken(token);
+    return staffService.findStaffByEmail(userInfo.get("email").asText());
```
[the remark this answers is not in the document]
emab: Yeah agreed. I wasn't sure whether methods like this should be in a seperate class, and have all methods that talk to Auth0 in one class and methods that use the information from Auth0 in another?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
emab: Yeah I wasn't thinking properly when I did this! We should definitely handle these errors in the controller.

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
```diff
@@ -1,51 +1,207 @@
...
+import org.springframework.boot.json.JsonParseException;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.stereotype.Service;
 
 /**
- * Dummy class which represents user authentication.
+ * Utils for handling authorisation tokens and general authentication related methods.
  */
+@Service
 public class AuthenticationUtils {
```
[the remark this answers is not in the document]
emab: Yeah agreed. I wasn't sure whether methods like this should be in a seperate class, and have all methods that talk to Auth0 in one class and methods that use the information from Auth0 in another?

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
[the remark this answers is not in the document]
emab: Yeah agreed. I wasn't sure whether methods like this should be in a seperate class, and have all methods that talk to Auth0 in one class and methods that use the information from Auth0 in another?

backend/src/main/java/com/rota/api/StaffService.java
emab: Yep this sounds good, I'll make the change!

backend/src/main/java/com/rota/auth/AuthenticationUtils.java
emab: We need this method to get the actual userInfo json, and then use the other utility methods to get bits of information from the json currently.

floyd924:
I am going to check this branch out locally and see what else needs to be done this morning

floyd924:
To those reading today, this is still in development so we do not yet require another review.

After discussions today, we still want to:
 
- Add some more meaningful error messages
- Use server-side roles to establish permissions, so remove and code we don't need which would get that info from AuthO
- update the front-end to attach a valid token to every request made
- tidy up

floyd924:
So at present:

 You can now add an entry into data.sql with an email address matching your email address used to obtain your token. Then when you hit an endpoint such as 'myShifts' the email is extracted from the access token and shifts returned for that user.

The admin check is currently still performed within the classes as before. I'd like to be able to do this in the SecurityConfig (hence the comment line), but this will require the addition of an additional JwtTokenFilter class such as this one: https://github.com/ScottLogic/mtg-app/blob/master/cardstore/src/main/java/com/scottlogic/cardstore/authentication/JwtTokenFilter.java


Next, a valid token is currently hard coded in 'global.ts' on the front-end. This should be removed, and a fresh token attached to each outgoing request. This should be done in a request handler class such as 'apiService', which contains basic generic 'get' and 'put' methods, which take a url as an argument, then attach all necessary headers.

PR still contains some experimental code. Can be cleaned up now we know what we do/don't need.


**In summary, tasks to do are:**

- append a fresh token to all front-end requests rather than a hard-coded value
- add a JwtTokenFilter class to the back end so that we can perform an admin check in SecurityConfig rather than manually extracting the role
- Remove back-end code we've decided not to use, as there's a bit left over from experimental setups

emab:
> **In summary, tasks to do are:**
> 
> * append a fresh token to all front-end requests rather than a hard-coded value
> * add a JwtTokenFilter class to the back end so that we can perform an admin check in SecurityConfig rather than manually extracting the role
> * Remove back-end code we've decided not to use, as there's a bit left over from experimental setups

I've done my best to implement these:
- A fresh token is added to all requests, so a user must be logged in to send any request.
  - Added ApiService as suggested by Iain.
  - Added a temporary loading check to the `App` component which is needed to let Auth0 authenticate.
- A RoleFilter now checks a user role against our database using their email address. If they have insufficient role, they get a 403 currently. This can be changed using the `ExceptionHandler` in another PR I believe.
- I removed the back-end code which tried to use the Auth0 management API as it is no longer necessary. The only external call is one to the Auth0 `/userinfo` endpoint to get the users email address.

backend/src/main/java/com/rota/auth/SecurityConfig.java
```diff
@@ -0,0 +1,86 @@
...
+    return jwtDecoder;
+  }
+
+  @Override
+  protected void configure(HttpSecurity http) throws Exception {
+    http.cors().and()
+        .csrf().disable()
+        .addFilterBefore(roleFilter(), BasicAuthenticationFilter.class)
+        .authorizeRequests()
+        .antMatchers("/**/myShifts/**").hasAuthority(Role.USER.toString())
```
AdrianBorg, line 51: would a manager be able to access this endpoint?

frontend/src/service/apiService.ts
```diff
@@ -0,0 +1,27 @@
+export const get = async (url: String, token: String | undefined) => {
```
AdrianBorg, line 1: never thought of doing this but I find it pretty neat.

backend/src/main/java/com/rota/auth/SecurityConfig.java
emab: Ah no they wouldn't. I've just looked in the docs and I think I can use `.hasAnyAuthority(Role.USER.toString(), Role.MANAGER.toString())` to give both access. I'll test now.

AdrianBorg: That'd be good for now. 

Although eventually we'd probably want to set something up where we'd have a hierarchy of permissions, where a manager would be able to do everything a user can and more. But I think that should be its own separate issue/ticket.

frontend/src/service/apiService.ts
```diff
@@ -0,0 +1,27 @@
...
+  const response = await fetch(url as RequestInfo, {
+    headers: { Authorization: `Bearer ${token}` },
+  });
+  return response;
+};
```
tjanowsk, line 6: I'm going to be a pain in the neck, but what happens if the GET call fails for whatever reason?
I think we need a way to handle errors (which I haven't done previously myself, so who am I to criticise).
You can add `.catch(callbackFunction)` to `await fetch`, but then the error needs to be propagated out of the function.

There are two ways I can think of:
1. change the function signature to accept an error callback function that will tell you what to do if an error is encountered.
2. we can stick to the functional programming approach and use e.g. `Result` from `folktale` package https://folktale.origamitower.com/api/v2.3.0/en/folktale.result.html - the function will then return either `Ok(response)` or `Err(error)`.

The latter option looks nicer to me, but I'm happy with either - we do need some feedback about why a call fails either way.

frontend/src/components/loginOutButton/index.tsx
```diff
@@ -25,6 +31,14 @@ const LoginOutButton = () => {
         onClick={handleLoginOutButtonClick}>
         {isAuthenticated && !loading ? 'Log out' : 'Log in'}
       </Button>
+      <Button
+        variant="contained"
+        onClick={async () => {
+          const token = await getTokenSilently();
+          console.log(token);
```
tjanowsk, line 38: This `console.log` needs to go.

frontend/src/features/staffSlice.ts
```diff
@@ -28,22 +28,20 @@ export const staffSlice = createSlice({
...
-export const deleteStaff = (id: string) =>
-  fetchWithAuth(STAFF_DELETE_URL + id, 'PUT', getAuthenticationToken);
+export const deleteStaff = (
+  id: string,
+  token: String | undefined
+): AppThunk => (dispatch) =>
+  put(STAFF_DELETE_URL + id, token)
+    .then((response) => response.json())
+    .then((response) => dispatch(set(response)))
+    .catch((err) => console.error(err));
```
tjanowsk, line 44: I'm OK with `console.error`, but we can squelch the warning here by using `toastr.error` instead. The behaviour will be different and the user will get a much more noticeable feedback, so not sure if that's what we want.

frontend/src/service/apiService.ts
emab: I wasn't sure if it was best to have it here or in the thunk itself. I felt like I was repeating myself in `staffSlice`. You had a nice function in there before which could adapt to the different methods, I'm sure it could be neatened up a bit?

frontend/src/components/employees/index.tsx
```diff
@@ -159,8 +164,12 @@ const Employees = () => {
 
   // Fetch data when component loads
   useEffect(() => {
-    dispatch(fetchStaff);
-  }, [dispatch]);
+    const getStaff = async () => {
+      const accessToken = await getTokenSilently();
```
tjanowsk, line 168: Can this fail? If so we need error handling here as well.

backend/src/main/java/com/rota/database/orm/engagement/Engagement.java
```diff
@@ -33,5 +35,6 @@
   private double hoursWorked;
 
   @ManyToOne(fetch = FetchType.LAZY)
+  @OnDelete(action = OnDeleteAction.CASCADE)
```
tjanowsk, line 38: I thought we wanted to keep the engagements if the staff was removed? Then again, the staff is never removed from the database, but made inactive instead, so probably irrelevant.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -19,90 +20,72 @@
...
-    final Integer user =
-        AuthenticationUtils.getUserFromToken(authString)
-            .orElseThrow(
-                () -> new ResponseStatusException(HttpStatus.UNPROCESSABLE_ENTITY,
-                    "Invalid user ID format in the token.")
-            );
+    final JsonNode userInfo = authentication.getUserInfoFromToken();
+    final Staff staff = authentication.getUserFromJson(userInfo).orElseThrow(
+        () -> new ResponseStatusException(HttpStatus.NOT_FOUND, STAFF_NOT_FOUND_MESSAGE));
+    // put the above 2 lines in a new method?
```
tjanowsk, line 66: remove unneeded comment

frontend/src/components/employees/index.tsx
```diff
@@ -117,7 +119,9 @@ const renderTableRow = (tableColumns: TableColumn[]) => (row: StaffData) => (
 const renderTable = (tableColumns: TableColumn[]) => (data: StaffData[]) => (
   <Table size="small">
     {renderTableHeaders(tableColumns)}
-    <TableBody>{data.map(renderTableRow(tableColumns))}</TableBody>
+    <TableBody>
+      {data.length > 0 && data.map(renderTableRow(tableColumns))}
```
tjanowsk, line 123: Won't this give
`<TableBody>false</TableBody>` if data.length is 0?
`map` already handles empty lists just fine.

frontend/src/service/apiService.ts
```diff
@@ -0,0 +1,27 @@
...
+  const response = await fetch(url as RequestInfo, {
+    method: 'PUT',
+    headers: { Authorization: `Bearer ${token}` },
+  });
+  return response;
+};
+
+export const post = async (
+  url: String,
+  token: String | undefined,
```
tjanowsk, line 18: The syntactic sugar for optional arguments which you can try is
`token?: string`
Also, I thought `string` was lower-case in typescript. Is upper-case OK as well?

frontend/src/components/loginOutButton/index.tsx
emab: This is here for developers to easily get their access token for testing, it will most definitely be taken out in the future.

I can put some comments around it to make its purpose more obvious?

frontend/src/features/staffSlice.ts
emab: Yeah okay this is a better option I think. I'll use this for now which will at least avoid any warnings, we will just need to make sure that the error that is shown makes sense for the situation.

backend/src/main/java/com/rota/database/orm/engagement/Engagement.java
emab: Ah yeah I forgot that was a thing! The database tests contained a test where a staff member was removed and it was complaining about engagement foreign keys due to that staff member no longer existing.

Should I remove this and remove the test on removing staff?

tjanowsk: We should definitely discuss this in the call, I don't think our staff removal policy is 100% clear!

frontend/src/components/employees/index.tsx
emab: If it fails it returns an undefined token, so it will just form a get request without the correct authorisation. I'm not sure if/how this method will fail unless something else has gone very wrong at Auth0 end.

I could do a check for `accessToken` being undefined or not?

frontend/src/components/employees/index.tsx
emab: I didn't change this, must have been Iain?

Although I believe this notation will perform a check to see if there is any data. If there is no data, then it will not perform the map and show nothing. I think I've used this before to stop an error caused by data being undefined/empty.

frontend/src/service/apiService.ts
emab: After a bit of reading, you shouldn't use capitals as it refers to Object types rather than the primitive types or something like that. I'll change all of these.

I think JS is just being nice.

frontend/src/service/apiService.ts
tjanowsk: I think this is better the way you have it now. It allows us to use the api calls without invoking redux. There is some repetition both here and in staffSlice, but you can avoid it by writing some helper functions, e.g. here:
```
const fetchUrl = (method: string) => async (url: string, token?: string, body?: string) => 
    await fetch(url as RequestInfo, {
      headers: { Authorization: `Bearer ${token}` },
      method,
      body,
  });

export const get = fetchUrl('GET');
```

and in staffSlice by passing the function as an argument. 
It's a bit cleaner, but not a big deal imo.

frontend/src/components/loginOutButton/index.tsx
tjanowsk: We definitely don't want to forget removing this later, printing sensitive information to the console makes me :scream: .
I suggest removing this now and if someone needs to debug a token later, they'll be able to figure it out.
Alternatively, you can comment it out and add a comment why it's there.

emab: I moved it to a separate component in a Dev folder to take it out of any components which will be used in MVP. I'll remove it from the sidebar and then someone can add it in if they wish to use it.

frontend/src/components/employees/index.tsx
tjanowsk: OK, I've learnt something. This is a logical statement which returns `false` when data.length is 0, but if false is passed as an element's child in JSX it is displayed as an empty tag
https://react-cn.github.io/react/tips/false-in-jsx.html

I still don't see why this guard would be necessary, which edge case does it guard against?
`map` handles `[]` just fine and typescript should prevent any use of `undefined` or `null` (which will break anyway when you try to access the object's length).

We can leave it for now and ask Iain when he's back.

frontend/src/components/employees/index.tsx
tjanowsk: I think it's fine as it is, that was a genuine question :slightly_smiling_face:

frontend/src/components/employees/index.tsx
emab: Yeah I was thinking that - I've done a check before but just done, for example:

```{data && data.value}```

to ensure that data.value isn't displayed until `data` is present to prevent undefined errors. But yeah ok I'll leave it and we can discuss later.

File: frontend/src/service/apiService.ts (hunk 16 above)
Comment (emab) [reply frontend/src/service/apiService.ts]: What would happen in a case where body isn't needed for example in a GET request? Can body be left as undefined?

Comment (tjanowsk) [reply frontend/src/service/apiService.ts]: I think you can just omit it and it will be left as `undefined`. Typescript won't complain, because it's declared as an optional argument.
The only potential problem is if you want body but not the token, I think you'd have to do
`fetchUrl('url', undefined, 'token')`.

I think it should be fine, but it should be verified.

Comment (emab) [reply frontend/src/service/apiService.ts]: Ah yeah I see. I think for now I will leave it as it is and keep the method invocations simple but we can easily change this down the line!

File: backend/src/main/java/com/rota/database/orm/engagement/Engagement.java [hunk 24]
```diff
@@ -14,6 +14,8 @@
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import org.hibernate.annotations.OnDelete;
+import org.hibernate.annotations.OnDeleteAction;
```
tjanowsk, line 18: These imports are not used anymore, I think

emab: Surprised they didn't get picked up in format tests! All cleaned up now. 👍

emab:
I've just worked out how to get the users email included in the access token, meaning we don't have to send the access token to Auth0 to get the email in a get request. 👍

PR #71, opened 2020-04-23
feature 69 - create getStaff endpoint

tjanowsk:
/staff/get endpoint, returns all active staff members when queried by a manager.


backend/src/main/java/com/rota/api/dto/StaffDto.java
```diff
@@ -33,6 +41,25 @@
   @ApiModelProperty("Preferred working dates of the employee")
   String preferredDates;
 
-  @ApiModelProperty("Is the staff currently inactive?")
+  @ApiModelProperty("Is the staff currenlty inactive")
```
floyd924, line 44: typo in curently

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -65,37 +66,41 @@
     return staffService.getStaffEngagementsBetween(user, start, end);
   }
 
+  private void verifyManagementRole(String authString) {
```
emab, line 69: In a future PR this could also be moved to `AuthenticationUtils` or `StaffService`.

tjanowsk: I kept it here because all this function does is throw ResponseStatusException, which in my view is controller's responsibility.
This keeps AuthenticationUtils and StaffService independent of any http status.

emab: Yep cool!

PR #74, opened 2020-04-23
Feature/48 json server setup

aBluePawn:
Add the json-server to dev dependencies,
add some data to the json server db.
This is all new to me, I know I still have tones to learn, so any suggestions are welcomed.


frontend/src/json_server/db.json
```diff
@@ -0,0 +1,7476 @@
...
+    {
+      "first_name": "Johnny",
+      "surname": "Herbert",
+      "start_date": "2018-08-03",
+      "contracted_hours": 40,
+      "hourly_rate": 2,
+      "role": "USER",
+      "job_title": "barista",
+      "inactive": false,
+      "id": 6
```
tjanowsk, line 67: I don't suppose we need a metric ton of Johnny Herberts?

aBluePawn: no, we don't, my mistake. I had a test running and it kept adding the same person to the db, and I pushed while the server was running.
But I cannot edit this branch,
"You can’t commit to feature/48-json-server-setup because it is a protected branch."
What should I do?

emab: Can you run us through what commands you're using when creating and pushing your branch? It shouldn't be protected by default.

aBluePawn: ok, that solved it. Thanks

frontend/src/json_server/db.json
```diff
@@ -0,0 +1,106 @@
...
+    {
+      "start": "2018-08-05 12:00",
+      "end": "2018-08-05 17:00",
+      "type": "SHIFT",
+      "hours_worked": 3,
+      "staff_id": 4
+    },
+    {
+      "start": "2020-08-03 00:00",
+      "end": "2021-08-03 01:00",
```
floyd924, line 100: Is this such a long holiday on purpose?

aBluePawn: no, not on purpose. ( I just copied the data from the backend, data.sql file).
The idea is that as we define the responses expected form the endpoints, we will update the data in this file.

PR #75, opened 2020-04-27
Feature/58 employee list screen

tjanowsk:
Run the server from the backend directory using
`./gradlew bootRun`
Run the front end from the frontend directory using
`npm start`

The employees component reads the data from the server and renders an html table.
The 'Add', 'pencil' and 'X' buttons don't do anything yet, just print a message.
CSS still needs some work to make it look like the design.

This PR may include some recent commits into master, I'll have to rebase eventually.

Screenshot with plain CSS:
Employees

Edit: Screenshot with material UI:
employees


frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+  contractedHours,
+}: StaffData) => (contractedHours >= fullTimeHours ? 'full' : 'part');
+const editUserButton = (editFunction: (o: number) => () => any) => ({
+  id,
+}: StaffData) => (
+  <button
+    type="button"
+    className={employeesStyle.editButton}
+    onClick={editFunction(id)}>
+    &#x1F589;
```
emab, line 50: Here we could also make use of Material UI icons: https://material-ui.com/components/material-icons/

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -23,6 +24,7 @@
 import org.springframework.web.server.ResponseStatusException;
 
 @RestController
+@CrossOrigin(value = "*")
```
floyd924, line 27: I couldn't find which branch this was on, so wanted to flag it just incase. Something like this disables CORS, right? It strikes me as a bad idea to put this at the top of a controller class, as we'd need to do so on all controller classes we create, right?

I've seen cors disabled for a list of permitted URLs setup in a securityConfig file, similar to this one on Core Competition Engine on the link below. However, there are always multiple ways to do this, as magic did something different (but I cannot understand what....). Can you have a quick google and see if any solutions jump out?


https://github.com/ScottLogic/core-competition-engine/blob/e6d868dc64fbff06bbac9e153b9d6645759137b3/src/main/java/com/scottlogic/competitionEngine/authorization/TestSecurityConfig.java

tjanowsk:
> Looks good to me.
> 
> There's quite a lot of changes in parts of the code not related to the employe list screen such as the sidebar scss. Is it needed to make this part display correctly?

No, these changes were pulled from master. Basically I made a bit of a mess with this branch, first branching off of master and then realising I need the back-end stuff so rebased onto feature/69.
But feature/69 doesn't have the new changes from master yet and I can't be bothered to rebase everything :(

frontend/src/components/employees/employees.module.scss
```diff
@@ -0,0 +1,50 @@
...
+  margin: 0 1em;
+  grid-template:
+    'header addbutton' 5em
+    'table table' auto
+    / auto 10em;
+}
+
+h1 {
+  grid-area: header;
+  font-family: Arial, 'Helvetica Neue', Helvetica, sans-serif;
```
floyd924, line 13: Can you list this font in the `constants` file, perhaps as `$table_font` or something?

frontend/src/components/employees/employees.module.scss
```diff
@@ -0,0 +1,50 @@
...
+.removeButton {
+  color: red;
+  font-weight: 1000;
+}
+
+.editButton:hover,
+.removeButton:hover {
+  font-size: 1.2em;
+  margin: 0em;
+}
```
floyd924, line 50: I think you can put this inside the `.editButton, .removeButton` section on line 38, using the `&:hover{ // stuff here }`

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+import TableHead from '@material-ui/core/TableHead';
+import TableBody from '@material-ui/core/TableBody';
+import TableRow from '@material-ui/core/TableRow';
+import TableCell from '@material-ui/core/TableCell';
+import toastr from 'toastr';
+import { StaffData, fetchStaff, selectStaff } from '../../features/staffSlice';
+import 'toastr/build/toastr.min.css';
+import employeesStyle from './employees.module.scss';
+
+const STAFF_FETCH_URL = 'http://localhost:8080/staff/get';
```
floyd924, line 15: I think we should move this to staffSlice. What do you think?

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+  id,
+}: StaffData) => (
+  <button
+    type="button"
+    className={employeesStyle.editButton}
+    onClick={editFunction(id)}>
+    &#x1F589;
+  </button>
+);
+const removeUserButton = (removeFunction: (o: number) => () => any) => ({
```
floyd924, line 53: What would be your thoughts on having an extra line of space between functions? When functions span several lines it feels easier to read with extra space, though it may look unneccessary when functions are only 1 line.

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+);
+
+/**
+ * Specifies the list of table columns to be displayed.
+ * Each  column has the following properties:
+ * id - column's unique id (also its class name for css)
+ * name - column name - the top row of the table
+ * content - a function taking StaffData as an argument
+ *    and returning the value of the corresponding table
+ *    cell
```
floyd924, line 71: Can you make indentation consistent in this comment?

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+ */
+const TABLE_COLUMNS = [
+  { id: 'name', name: 'Name', content: getName },
+  { id: 'job', name: 'Job title', content: getJobTitle },
+  {
+    id: 'partfull',
```
floyd924, line 77: This is probably a linter splitting these lines? If not, Is it possible to be consistent - either have all object split over multiple lines, or none?

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+  </Box>
+);
+
+const Employees = () => {
+  const staffList = useSelector(selectStaff);
+  const dispatch = useDispatch();
+
+  // Fetch data when component loads
+  useEffect(() => {
+    dispatch(fetchStaff(AUTHENTICATION_TOKEN, STAFF_FETCH_URL));
```
floyd924, line 133: See my ealier comment about moving the url to the staffSlice class. 

The Auth_Token stuff will be refactored once we have the Auth PR merged

backend/src/main/java/com/rota/api/StaffController.java
emab: In a previous project I set up a CORS config file which allowed the front end API address only but in all rest controllers.

emab:
This is currently set to merge into feature/69... is that correct? Is it not meant to be merging into master? Or does it just depend on feature/69?

frontend/src/components/employees/index.tsx
tjanowsk: I put an extra tab to highlight that the last 2 lines refer to 'content'

frontend/src/components/employees/index.tsx
tjanowsk: That's prettier being weird. I think it tries to keep objects in line if they are short enough and splitting them up otherwise.

tjanowsk:
> This is currently set to merge into feature/69... is that correct? Is it not meant to be merging into master? Or does it just depend on feature/69?

This contains changes introduced in feature/69 so we shouldn't move this directly to master before feature/69 goes in.

frontend/src/components/employees/index.tsx
tjanowsk: We have a 'constants' directory, so I'd probably move it there.

frontend/src/components/employees/employees.module.scss
```diff
@@ -0,0 +1,50 @@
...
+  text-align: right;
+  width: 1em;
+}
+
+.editButton,
+.removeButton {
+  border: none;
+  outline: none;
+  background-color: transparent;
+  font-weight: 100;
```
QuarkNerd, line 36: font-weight seems like something we might reuse in other parts of the app for consistency? If so maybe we should move this to constants too?

frontend/src/components/employees/index.tsx
floyd924: Fair enough

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+  <Box display="flex" justifyContent="flex-end" alignItems="center">
+    <Button
+      className={employeesStyle.addButton}
+      color="primary"
+      variant="contained"
+      size="small"
+      onClick={addUser}>
+      Add
+    </Button>
+  </Box>
```
QuarkNerd, line 124: Question: What benefits does using Box give us over using a div?

frontend/src/components/employees/index.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { useEffect } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import Box from '@material-ui/core/Box';
+import Button from '@material-ui/core/Button';
+import Table from '@material-ui/core/Table';
+import TableHead from '@material-ui/core/TableHead';
+import TableBody from '@material-ui/core/TableBody';
+import TableRow from '@material-ui/core/TableRow';
+import TableCell from '@material-ui/core/TableCell';
```
QuarkNerd, line 9: These imports can be changed to` import {TableCell, TableRow, TableBody, TableHead, Table, Button, Box} from '@material-ui/core';` you can still have each element on a new line. What I am suggesting is what I am used to, not sure if it is the accepted

tjanowsk: What I'm worried about is that this would import the entire `@material-ui/core` into the bundle, including elements we don't need. I'm not sure how clever webpack is about this, but this is how examples on material ui website were structured.

frontend/src/components/employees/employees.module.scss
tjanowsk: This one is there to make the pencil icon look right. If we choose to use material UI icons it will no longer be needed.

frontend/src/components/employees/index.tsx
QuarkNerd: Ah, fair eniough

emab:
If this PR is rebased will that remove some of the code that has already been added? Which bits need reviewing at the moment?

tjanowsk:
> If this PR is rebased will that remove some of the code that has already been added? Which bits need reviewing at the moment?

Depends how you do the rebase, I guess. I keep all the commits, so everything should still be there.
The new part since last PR is everything in `backend` directory, but just to be on the safe side, I'd say everything needs reviewing :slightly_smiling_face: 
I know it's a pretty big PR, I'll try to keep them smaller in the future.

frontend/src/components/employees/index.tsx
emab: https://material-ui.com/guides/minimizing-bundle-size/

This explains why this is happening, and your way of doing it is a solution. I think it will only happen during the development build.

Option 2 provides a solution using babel plugins which will keep the bundle size small on developer build and allow the nice new imports.

frontend/src/components/employees/index.tsx
```diff
@@ -1,11 +1,157 @@
...
+const getName = ({ firstName, surname }: StaffData) =>
+  `${firstName} ${surname}`;
+
+const getJobTitle = ({ jobTitle }: StaffData) => jobTitle;
+
+const partFullTime = (fullTimeHours: number) => ({
+  contractedHours,
+}: StaffData) => (contractedHours >= fullTimeHours ? 'full' : 'part');
+
+const editUserButton = (editFunction: (o: number) => () => any) => ({
```
emab, line 53: I'm pretty new to TS and new JavaScript functions... can you explain this line to me please? 😊

tjanowsk: It defined a new javascript function which takes a single argument, which is a function 
that takes a number as an argument and returns a function with no arguments and has any return type. Sounds confusing, I know, but it's simple once you're used to it.

If you have a function
```
const f = x1 => x2 => [etc.] => xn => result 
```
it is a 'curried' function where you can think of x1,...,xn as arguments and the part after the last `=>` as the result.
You can provide all arguments immediately `f(x1)(x2)[etc.](xn)` and get the result, or you can apply it partially `f(x1)(x2)[etc.](x(n-k))` and get a curried function from `x(n-k+1)=>...=>xn=>result`

floyd924:
1 more thing I'd like to flag up before this goes in! I see you've used the type `any` on a couple of occasions. I can't remember if it's already been discussed - can we change this to something more meaningful? Otherwise there's not much point in us using typescript over plain JS :)

PR #76, opened 2020-04-29
Hotfix/Login Button

emab:
## Hotfix
Login button does not work in current master branch. This pull request fixes it and updates its behaviour to work as expected:
- When you log in, you are redirected to the Route you logged in from
- The `/` route still redirects without causing login issues

### Issue

This bit of code is causing issues with Auth0:

```
<Route exact path="/">
  <Redirect to="/employees" />
</Route>
```

The issue was that the original `callback_uri` was`http://localhost:3000`, but when that route was reached you were instantly redirected to `/employees`. To fix this, I updated the `callback_uri` to `http://localhost:3000/callback` which is a standard way to handle this.

This then caused the issue that the user would be returned to `http://localhost:3000` by default after login, which would then always try and redirect to `/employees` again. 

To fix this, `BrowserRouter` was replaced with `Router`. This is required so that the custom `history` object can be used for routing. The custom history object is passed to the `Router` as a prop and it then allows `onRedirectCallback()` in `index.tsx` to push the `targetUrl` defined by the login button to the history object, which automatically navigates to the page that you logged in from.

### DIFF
--- a/frontend/src/components/home/index.tsx
+++ b/frontend/src/components/home/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Route, Redirect, BrowserRouter as Router } from 'react-router-dom';
+import { Route, BrowserRouter as Router } from 'react-router-dom';
 import Sidebar from '../sidebar';
 import Employees from '../employees';
 import Rota from '../rota';
@@ -14,9 +14,6 @@ const Home = () => (
       <Route path="/summary" component={Summary} />
       <Route path="/" component={Sidebar} />
     </div>
-    <Route exact path="/">
-      <Redirect to="/employees" />
-    </Route>
   </Router>
 );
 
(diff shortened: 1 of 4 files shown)


frontend/src/components/home/index.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { Route, Redirect } from 'react-router-dom';
 import Sidebar from '../sidebar';
 import Employees from '../employees';
 import Rota from '../rota';
 import Summary from '../summary';
 import homeStyles from './home.module.scss';
 
 const Home = () => (
-  <Router>
+  <>
```
floyd924, line 10: I've been urged to avoid creating containers using an empty tag. I cannot find much info about why you should or shouldn't at the moment, but I'd probably just make this a div with a, informative className so we can identify it later on

emab: I could potentially move the Router back here?

Issue #78, opened 2020-04-29
BUG - prettier comands not working


QuarkNerd:
On my shell Windows10- git bash the prettier commands no longer work.

 I believe this is due to the change brought in by https://github.com/Furlough-Friends/rota-flex-101/pull/63

PR #79, opened 2020-04-29
Hotfix/prettier and linter

QuarkNerd:
https://github.com/Furlough-Friends/rota-flex-101/issues/78

Originally the lint and prettier commands had no quotes around file path, however this caused @tjanowsk issues so single quotes were added (https://github.com/Furlough-Friends/rota-flex-101/pull/63) . This is now causing me issues (I am not blaming)  and I have shifted to using \". I believe this should work on everyone's system


emab:
I can confirm that `npm run format` and `npm run lint` both work using this branch on Windows 10 from Cmder.

emab:
Can also confirm mine do not currently work on `master` branch.

emab:
Gurveer which environment are you developing on? And which is @tjanowsk using? If this works for all three of us can we get this change into master?

tjanowsk:
> Gurveer which environment are you developing on? And which is @tjanowsk using? If this works for all three of us can we get this change into master?

Works for me. The problem is that I was playing around too much.
I typically use `zsh` shell (which is similar to bash but better in every conceivable way),
but recently I wanted to try out `fish`, which is the `f`riendly `i`nteractive `sh`ell, but it's not POSIX compliant so I suspect that this causes some issues.
It's a nice shell otherwise.

Issue #81, opened 2020-04-29
Refactor some sections of the Create User endpoint


floyd924:
#66 was blocking other tickets so was merged into master with a couple of outstanding changes still to be made.

Some of these changes may have been made in a later PR. Check the current state of master and other pending PR's, and ensure the code has been refactored where required:

api/StaffController line 94:
There is a check for role != manager - we should instead call a `hasManagerPermissions` method

api.staffService line 37:
Refactor the check for engagement start time to be `!engagement.getStart().isBefore(startTime)`

api/dto/form/CreateStaffForm - how similar is this class to the StaffDTO class? If there is enough overlap, does it make sense to get rid of 1?

auth/authenticationUtils line 42:
There is currently a check to see if the user has 2 roles. This could be refactored to check the arrayList of roles contains 'manager', rather than just checking the length of the array = 2.

emab:
> auth/authenticationUtils line 42:
There is currently a check to see if the user has 2 roles. This could be refactored to check the arrayList of roles contains 'manager', rather than just checking the length of the array = 2.

It's currently checking if the length of the token is even and then returning an optional of a role based on that alone for development purposes. I'll change this in a future PR when I properly integrate the backend with Auth0.

Can that be left as it is for now?

tjanowsk:
> api/dto/form/CreateStaffForm - how similar is this class to the StaffDTO class? If there is enough overlap, does it make sense to get rid of 1?

For better or worse, I have removed form/CreateStaffForm.java as a part of feature/69 PR, its functions, including input validation, have been moved to StaffDto.java

I don't feel strongly about my other 2 comments, but I could include them in feature/69 as well?

emab:
I've also done that in the refactor pull request @tjanowsk

PR #82, opened 2020-04-29
Refactor/81 create user endpoint

emab:
This addresses the refactors mentioned in #81.

The only refactor I didn't do is the last one, as that method will be changed completely in a future PR when Auth0 is properly linked to the backend, as the users role is stored there and can be accessed via an Auth0 Management API call.

#### Update 01/05
This is now ready to merge, and is pending approval again after resolving conflicts.


backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -34,7 +38,7 @@
     return engagementRepository.findByStaffId(staffId).stream()
         .filter(engagement ->
             (engagement.getStart().equals(start)
-                || engagement.getStart().isAfter(startTime))
+                || !engagement.getStart().isBefore(startTime))
```
tjanowsk, line 41: With this `engagement.getStart().equals(start)` is no longer needed.
These two lines can be written simply as `!engagement.getStart().isBefore(startTime))`

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -34,7 +38,7 @@
...
           && (engagement.getEnd().equals(endTime)
                 || engagement.getEnd().isBefore(endTime)))
```
tjanowsk, line 43:
```
&& !engagement.getEnd().isAfter(endTime)
```

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -50,4 +54,19 @@
...
+   * @return true if user has manager permissions.
+   */
+  public boolean hasManagerPermissions(String authString) {
+    final Role role = AuthenticationUtils
+        .getUserRoleFromToken(authString)
+        .orElseThrow(() ->
+            new ResponseStatusException(HttpStatus.UNAUTHORIZED, "Failed to authorize.")
+        );
+    return role == Role.MANAGER;
+  }
```
tjanowsk, line 71: I would move this to AuthenticationUtils.java

backend/src/main/java/com/rota/api/StaffService.java
emab: Ah yep I see good point, I'll update this

backend/src/main/java/com/rota/api/StaffService.java
emab: I was thinking this, but wasn't sure if it belonged there or not. I guess since it still takes an `authString` as a parameter it should all be handled in `AuthUtils`.

emab: Should the `StaffService` still contain a `hasManagerPermissions()` method? Because of the way spring security handles an auth session, we can access the access token for the current execution thread at any time by using:

```SecurityContextHolder.getContext().getAuthentication()```

This could be done in `AuthenticationUtils`.

tjanowsk: I don't feel very strongly about this. 
The way the classes work in my head is as follows:
AuthenticationUtils - all authentication/permission related methods
StaffController - all http REST-related methods
StaffService - helper methods for StaffController, reading from the database

This particular method is a mixture of AuthenticationUtils (`return role == Role.MANAGER`) and StaffController (`ResponseStatusException`), so I'd split it in 2 and put them in their respective areas.

I'm not too bothered about the way it's written now though.

emab:
Review approval required again after sorting merge conflicts.

PR #84, opened 2020-05-04
Feature/80 create page not found

emab:
## Summary
- Add a basic PageNotFound component. This is currently set up to display whenever a page is routed to that is not defined as a route in the `Home` component.
  - Has some basic styling.
  - Some styling was done in the component class. I didn't like this at first, but think it may be a limitation when using some Material UI components. If someone else can work out how to use CSS module to replace having to use `style={...}` that could be useful.
- Added Roboto font as recommended by Material UI.
- Made some changes in `Home` component:
  - I pulled the `Sidebar` from the router, as it is always displayed on the page.
  - I moved the other `Routes` into a `Switch`. This is so that only one is selected and displayed. 
  - I added a blank route to `/callback` to prevent the error page displaying when the user logs in.

Here is the not found page:

image


frontend/src/components/pageNotFound/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+const PageNotFound = () => {
+  const location = useLocation();
+  return (
+    <div className={errorPageStyle.errorPage}>
+      <Grid
+        container
+        alignItems="center"
+        justify="center"
+        style={{ minHeight: '100vh' }}>
+        <Box>
```
tjanowsk, line 16: I'm not very familiar with material UI, do we need both Grid and Box here?
If it's used for the alignment, can we just use Box? 
Can we use div and move the styling to css file (with `display: flex` etc.)?

emab: I've had a play around and nope I don't think we do. I was following some stack overflow advice on Material UI components. I've also not used it before - it's slightly different from Semantic UI which I have used before.

frontend/src/components/pageNotFound/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const PageNotFound = () => {
+  const location = useLocation();
+  return (
+    <div className={errorPageStyle.errorPage}>
+      <Box>
+        <h1>
+          <CancelIcon
+            className={errorPageStyle.icon}
+            style={{ fontSize: '90px' }}
```
floyd924, line 15: Is it possible to move this styling into the css file?

frontend/src/components/pageNotFound/pageNotFound.module.scss
```diff
@@ -0,0 +1,35 @@
...
+  grid-area: main;
+  width: 100%;
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+body {
+  background-color: #e9e9e9;
```
floyd924, line 11: Is this a particularly nice colour we might use elsewhere in the app? If so, maybe make it a variable

emab: I'm not even sure if it is to be honest! I'll have a look and see if we even need a bg colour here.

emab: Removed bg colour for now. We can add one later if needed.

PR #85, opened 2020-05-04
Feature/68 create delete modal

tjanowsk:
Delete user modal fully working now.

deletemodal


frontend/src/components/employees/index.tsx
```diff
@@ -136,9 +141,65 @@ const addButton = (
...
+      <Button
+        variant="contained"
+        color="secondary"
+        onClick={deleteCallback(staff)}>
+        DELETE
+      </Button>
+    </div>
+  </div>
+);
+
```
QuarkNerd, line 177: is it worth taking the modal out of this file and make it's own thing? I feel we are making this file to cluttered.

frontend/src/components/employees/index.tsx
```diff
@@ -136,9 +141,65 @@ const addButton = (
...
 const Employees = () => {
   const staffList = useSelector(selectStaff);
   const dispatch = useDispatch();
+  const [isModalOpen, setModalState] = useState(false);
+  const [modalBody, setModalBody] = useState(<div />);
+
```
QuarkNerd, line 183: is setting HTML elements in state a normal thing to do? My understanding was that state should aim to contain the minimum information possible to tell you (along with props) what the component should look like.

tjanowsk: I don't know what's normal, but in this case the modal needs to know which button in the table was clicked and this is the first solution that came to mind - setting the body in the callback function.

frontend/src/components/employees/index.tsx
tjanowsk: done

frontend/src/components/employees/index.tsx
```diff
@@ -139,6 +141,25 @@ const addButton = (
...
+  const closeModal = () => {
+    setDeleteModalState(false);
+  };
+
+  const openDeleteModal = (staff: StaffData) => () => {
+    setDeleteModalState(true);
+    setSelectedStaff(staff);
+  };
+
+  const tableColumns = makeTableTolumns(editUser, openDeleteModal);
```
QuarkNerd, line 162: typo, `makeTableTolumns` should have columns

tjanowsk: 'Tolumn' is the official word for 'table column', didn't you know? :stuck_out_tongue_winking_eye: 

fixed

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -123,4 +123,27 @@ private void verifyManagementRole(String authString) {
     verifyManagementRole(authString);
     return staffService.getActiveStaff();
   }
+
```
emab, line 126: After the backend PR is in (if it goes in current state) you no longer need to get the `AUTHORIZATION_HEADER`.

You can just call `verifyManagementRole()` and it will check the roles of the authorisation used in the current execution thread! 

Just a heads up of how it is currently working!

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -80,4 +81,17 @@ public boolean hasManagerPermissions(String authString) {
         .filter(staff -> !staff.isInactive())
         .collect(Collectors.toList());
   }
+
```
emab, line 84: Should we have a differentiation between deactivate and remove? Removing a staff member for me sounds like a permanent removal, and this is just deactivating. How would a manager permanently remove a member of staff, or is that not something they would ever do?

frontend/src/react-auth0-spa.tsx
```diff
@@ -81,7 +83,7 @@ export const Auth0Provider = ({
     try {
       await auth0Client!.loginWithPopup(o);
     } catch (error) {
-      console.error(error);
+      toastr.error(error);
     } finally {
```
emab, line 87: 👍 Good idea!

frontend/src/components/employees/DeleteModal.tsx
```diff
@@ -0,0 +1,67 @@
...
+  closeModalFunction: () => void;
+}
+
+const DeleteModal = ({ isOpen, staff, closeModalFunction }: Props) => {
+  const dispatch = useDispatch();
+
+  const deleteUser = ({ id, firstName, surname }: StaffData) => () => {
+    dispatch(
+      fetchStaff(getAuthenticationToken(), `${URL}/staff/remove?id=${id}`)
+    );
```
emab, line 28: Is it worth having a separate thunk for deleting a user? Or is it best to have one which takes parameters like this and can be used for different purposes?

QuarkNerd: I agree that they should be separate. It feels strange for a a function whos job is to fetch something specific to accept a URL

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: I thought the idea was that remove=deactivate in our case, so we get to keep all the user's past engagements intact.
Correct me if I'm wrong.

frontend/src/components/employees/DeleteModal.tsx
tjanowsk: I can make the change suggested under #89 here if this works for you.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -123,4 +123,27 @@ private void verifyManagementRole(String authString) {
...
+  /**
+   * Endpoint for the manager to remove a staff member by a given id.
+   * 
+   * @param authString manager's authentication token.
+   * @param id staff id to remove.
+   * @return An updated list of active staff members.
+   */
+  @GetMapping("/staff/remove")
```
floyd924, line 134: If the purpose of this endpoint is to update an object (change status to 'inactive') then this should be a PUT, rather than a GET.

frontend/src/components/employees/DeleteModal.tsx
floyd924: I'm a little confused here. We're calling a method called fetchStaff and passing it a url of our choosing, which could in theory not be related to a staff endpoint at all? I think we should refactor that to be staff specific, or rename it.

One possible solution would be to create a generic 'apiService' class which has a 'get' method on it, which takes a url (which should be a const somewhere), and appends any headers required. This will also make it easier to append a token to every single request later on.

per my earlier point about the endpoint being a PUT rather than a GET, we'll presumably need another method here anyway, rather than calling fetchStaff.

Let me know your thoughts on this 👍🏻 I've shared my thoughts on #89 as well

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: fixed

frontend/src/components/employees/DeleteModal.tsx
tjanowsk: I admit, the naming was awful. I have refactored it now, although rather than a class, I have a bunch of functions.
What would the fields of this apiService class? Just the base URL?

floyd924: Functions are probably fine for now, and we could think about a class over in the AuthZ PR that's open.

I haven't given it much thought tbh - my basic thinking was having a class similar to this. Particularly note lines 38, and 62, which are our basic methods:

https://github.com/ScottLogic/mtg-app/blob/master/client/src/shared/network/NetworkService.ts

frontend/src/components/employees/employees.module.scss
```diff
@@ -49,3 +49,30 @@ Table {
...
+  left: 50%;
+  transform: translate(-50%, -50%);
+  padding: 1em;
+  background-color: white;
+}
+
+.modalDeleteButtons {
+  display: flex;
+  justify-content: flex-end;
+  & Button {
```
floyd924, line 65: when nesting SCSS I don't think we need the `&` here, or line 73. Correct me if I'm wrong plz

tjanowsk: You're right, they don't seem to be needed.
Removed now.

floyd924:
I thought I posted this but must have got lost somewhere.....

What do you think about extending the space between the 'are you sure you want to delete' text and the cancel/confirm buttons? Seems to be a nice amount of padding at the top, bottom, and above the line, but slightly smaller in this space?

Screenshot 2020-05-06 at 11 11 38

frontend/src/components/employees/index.tsx
```diff
@@ -139,17 +137,41 @@ const addButton = (
...
+    setDeleteModalState(true);
+    setSelectedStaff(staff);
+  };
+
+  const tableColumns = makeTableColumns(editUser, openDeleteModal);
 
   // Fetch data when component loads
   useEffect(() => {
-    dispatch(fetchStaff(getAuthenticationToken(), STAFF_FETCH_URL));
+    dispatch(fetchStaff(getAuthenticationToken()));
```
QuarkNerd, line 162: Okay so this might not be a thing we have to fix in this PR but I also think that the frontend components shouldn't care about authentication if that makes sense? given many api calls will be using auth. The auth logic should be handled in one place where we define our api calls? Don't feel the need to change it here but something to consider

floyd924: That is something I think we can address as part of #67 so I think you can leave this just now

tjanowsk: eh, it's a quick fix so I just did it

PR #90, opened 2020-05-05
Feature/53 make sidebar collapsable - setup (part 1)

AdrianBorg:
Small PR where I saw some opportunities for improvement before beginning the collapsible part.

Changes done:
* Rearranged the size (width) styling for the grid layout and width of sidebar so that its easier to make the left panel expand to fill the space. 
* Allowing the sidebar to have a max width, below which it would be 25% of the view until a min width of the content.

* Substituted the use of Buttons and Link components in the sidebar with NavLink. NavLink creates `<a>` elements which can still be clicked and can be tabbed through. Navlink also allows to easily set an active style if we are at that location.


frontend/src/components/sidebar/index.tsx
```diff
@@ -11,23 +11,18 @@ const SidebarButton = ({
...
-      `/${endpoint}` === useLocation().pathname
-        ? sidebarStyles.selectedMenuOption
-        : ''
-    }`}
-    key={name}>
+  <NavLink
+    key={name}
+    to={`/${endpoint}`}
+    className={`${sidebarStyles.menuOption}`}
+    activeClassName={`${sidebarStyles.selectedMenuOption}`}>
```
tjanowsk, line 18: Neat, I didn't realise NavLink was a thing!

frontend/src/components/home/home.module.scss
```diff
@@ -1,6 +1,5 @@
 .home {
   display: grid;
-  grid-template:
-    'sidebar main'
-    /25% auto;
+  grid-template-areas: 'sidebar main';
+  grid-template-columns: auto 1fr;
```
tjanowsk, line 4: Instead of having sidebar width to auto and setting min and max width with sidebar's css, we can use the css minmax here directly
`grid-template-columns: minmax(200px, max-content) 1fr;`

AdrianBorg: I think I'd tried that but there are issues when the width would be less than 200px. I had gone through a few iterations but the minmax stuff works a bit weirdly where if max-content < 200px, it ignores the max part and keeps it as 200px always. 

Switching them over (to `minmax(max-content, 200px) 1fr` ) doesn't work as intended either, although it was the closest I got. When width was between max-content and 200, the sidebar was about 50% of the view, not 25%. And messing around with nested minmax statements wasn't too helpful to me (probably because it got more confusing).

I'd spent some time trying to mess around with it, trying to keep within the grid style stuff, and with everything I tried other than `auto`, the `sidebar` grid cell never properly resized. (having a max width, min width and 25% width in between).

emab:
Material UI has some [in-built support for drawers](https://material-ui.com/components/drawers/) if you're interested in having a look at that!

PR #96, opened 2020-05-07
Feature/34 Update staff endpoint with ExceptionHandler

emab:
### Staff endpoint and business logic
- Added a PUT endpoint to `/staff/{id}` which is used to update the staff member with that ID with a new StaffDto object.
- Added `updateStaff()` method in `StaffService`.
  - Checks to see if a staff member with that ID is in the database.
    - If not, throw `StaffNotFoundException`.
  - If that staff member exists, convert `StaffDto` to `Staff`, and set its ID to corresponding staff ID.
  - Save in database.
- Added `updateStaff()` tests.
### ExceptionHandler
Since a custom exception is thrown in the new code, I've added an exception handler for the controller layer. In this PR, if a `StaffNotFoundException` is thrown, it creates a nice error response message to be sent to the user. In this case it sends this response when the ID isn't found when updating a staff member:

image

This exception handler can be used to handle any exceptions that are thrown in the controller layer.

### Potential changes
In the `StaffController` we currently get all staff by making a GET request to `/staff/get`. Would it make more sense for it to just do GET `/staff` to return all staff? Then GET `/staff/$id` to get a specific staff member?

### DIFF
--- a/backend/src/main/java/com/rota/api/StaffController.java
+++ b/backend/src/main/java/com/rota/api/StaffController.java
@@ -15,6 +15,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -132,10 +133,10 @@ public class StaffController {
    * @param updatedStaff Updated staff information.
    * @return The updated {@link Staff} member.
    */
-  @PutMapping("/staff/$id")
+  @PutMapping("/staff/{id}")
   @ApiOperation(value = "Lets an authenticated manager update a staff member")
   public Staff updateStaff(
-      @RequestParam
+      @PathVariable
       @ApiParam(value = "Staff ID")
           int id,
       @RequestBody
--- a/backend/src/main/java/com/rota/api/StaffController.java
+++ b/backend/src/main/java/com/rota/api/StaffController.java
@@ -8,11 +8,9 @@ import com.rota.database.orm.staff.Staff;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
-
 import java.time.Instant;
 import java.util.List;
 import javax.validation.Valid;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -94,4 +94,20 @@ public void removeStaff(int id) {
...
+  /**
+   * Updates a staff member in the database.
+   *
+   * @param id           Of the staff member being updated.
+   * @param updatedStaff Updated staff details.
+   * @return The updated {@link Staff} member.
+   */
+  public Staff updateStaff(int id, StaffDto updatedStaff) {
+    // Check to see if that staff member exists, throw exception if not.
+    staffRepository.findById(id).orElseThrow(StaffNotFoundException::new);
```
tjanowsk, line 107: What happens when this exception is thrown? Does it crash the server?

emab: StaffNotFoundException is a RuntimeException, so it can be handled in an exception handler and respond with a 404 and a nice message at some point.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -125,23 +125,44 @@ private void verifyManagementRole(String authString) {
...
+   */
+  @PutMapping("/staff/$id")
+  @ApiOperation(value = "Lets an authenticated manager update a staff member")
+  public Staff updateStaff(
+      @RequestParam
+      @ApiParam(value = "Staff ID")
+          int id,
+      @RequestBody
+      @ApiParam(value = "Updated staff object")
+          StaffDto updatedStaff
```
tjanowsk, line 143: I think the syntax here is wrong - does it work for you?
I think what you want is
`@PutMapping("/staff/{id}"`
and `int id` should have `@PathVariable(name="id")` instead of `RequestParam`.

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: OK, for the moment it gives 500 - internal server error, which is not very helpful.
A possible solution would be to throw the error and have a try-catch block inside the controller, but I'm curious to see how your proposed solution works.

emab: I've updated - now throws a 400 with a nice error message.

backend/src/main/java/com/rota/api/StaffController.java
emab: It seemed to work when I tested it, but I agree that is the correct way of doing it! I'll update now.

tjanowsk: OK, weird, it gave me 404 every time I tried it.

emab: I swear I updated one but oh well! It's correct now so hopefully will work!

Issue #59, opened 2020-04-17
Create 'Add' user modal


smangan360:
**Requires #58**
When the user clicks the add button, a modal should appear with a form with input values with the required fields : 
- First Name 
- Surname
- Job title
- Contracted hours
- Pay
- Preferred days (radio button)

Should have three buttons: 
- Add - button should have same styling as initial 'Add' button at the base of the modal
- Cancel - button should be coloured slightly differently to distinguish a difference between functionality at the base of the modal beside the 'Add' button
- x - An close option in the top right of the modal 
**When the initial 'Add' button is clicked, the modal appears and there is a dark semi transparent overlay behind the modal allowing for more focus on the modal.**

Each input box should have two states: 
- Inactive - grey outline
- Active - when a manager is interacting with an input box the outline should change to a blue colour: 
![image](https://user-images.githubusercontent.com/54022972/79996005-d8395280-84af-11ea-9c0d-c154accc02e1.png)

emab:
Material UIs text inputs look a lot like how you want:

![image](https://user-images.githubusercontent.com/8884999/80085599-56e5cc80-8550-11ea-9de6-8b240c24b2e3.png)

And it's got cool built in things like this:

![image](https://user-images.githubusercontent.com/8884999/80085661-7250d780-8550-11ea-9eb8-a00a24b0e95c.png)


https://codesandbox.io/s/material-demo-m0tmb

QuarkNerd:
is pay hourly/weekly/monthly or do we give a choice?

QuarkNerd:
Do we want to add start date and ROLE?

smangan360:
#121 outlines that there can't be more than one users linked to an email address. It will probably have implications on the add modal for error handling. I'll add another ticket in now

PR #99, opened 2020-05-07
Feature/59 add user modal

QuarkNerd:
https://github.com/Furlough-Friends/rota-flex-101/issues/59

creating a modal to add users.

-- For now im switching the modal logic to use this https://stackoverflow.com/questions/35623656/how-can-i-display-a-modal-dialog-in-redux-that-performs-asynchronous-actions/35641680#35641680 which was recommended by a friend.

The benefits are that the modals are centrally controlled so we can easily create modals from any component. And we don't have to declare multiple modal components in any JSX


frontend/src/components/rootModal/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import DeleteUserModal from '../deleteUserModal';
+import {
+  ModalState,
+  selectModalDetails,
+  hideModal,
+} from '../../features/modalSlice';
+
+interface ModalHash {
+  [key: string]: any;
+}
```
QuarkNerd, line 13: if anyone can tell me the type for a generic react component, i will update this, but I need this to stop an error in line 22

tjanowsk: `JSX.Element` ?

frontend/src/components/deleteUserModal/index.tsx
```diff
@@ -10,15 +10,14 @@ import { StaffData } from '../../constants/employees';
 
 import { deleteStaff } from '../../features/staffSlice';
 import 'toastr/build/toastr.min.css';
-import employeesStyle from './employees.module.scss';
+import deleteUserModalStyle from './deleteUserModal.module.scss';
 
 interface Props {
-  isOpen: boolean;
   staff: StaffData;
   closeModalFunction: () => void;
```
tjanowsk, line 17: I believe closeModalFunction is not used anymore, because modal's state is handled by redux?

frontend/src/features/modalSlice.ts
```diff
@@ -0,0 +1,35 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
```
AdrianBorg, line 1: Haven't gone through this too finely yet, but just been going through the current PRs because I found (what I think at least) is an error in how we're using slices (if we're using duck folder structure).

I'm pretty sure that a slice is meant to be in the same component file that it'll be used with.
So this should go in the `employees` directory. Or the `rootModal` one, not really sure tbh.

That way we'd have related stuff in one folder.

frontend/src/features/modalSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+  modalType: string | null;
+  modalProps: Object;
+}
+
+const initialState: ModalState = {
+  modalType: null,
+  modalProps: {},
+};
+
+export const modalSlice = createSlice({
```
AdrianBorg, line 14: Also not sure if using the redux store to handle opening/closing the modal is best, unless we need other components to know that the modal is open/closed. 

I've been told that if its only that component that needs to know then just use react state. Which would be stored in the parent the modal, the employees page in this case.

QuarkNerd: So this method was actually suggested by my REACT mentor from my training. 
It also feels like the employee component is handling a lot of things rn, which isn't ideal. 

This method minimises work in the future too. I also imagine its quite likely we will get modals in someway coming up in other compoents, such as redifinng or deleting someones shifts?

frontend/src/features/modalSlice.ts
QuarkNerd: Okay, so this I guess depends on the exact folder structure we use. In my grad project we had components separate from redux store stuff. Maybe this calls fro more discussion

QuarkNerd:
There seems to be fields in the api to create users that were not in the ticket. is there a reason for this?
Namely start date and ROLE?

tjanowsk:
> There seems to be fields in the api to create users that were not in the ticket. is there a reason for this?
> Namely start date and ROLE?

Api fields correspond to the fields in the database, which include start date and role.

QuarkNerd:
> > There seems to be fields in the api to create users that were not in the ticket. is there a reason for this?
> > Namely start date and ROLE?
> 
> Api fields correspond to the fields in the database, which include start date and role.

So should they be added to the modal?

tjanowsk:
> > > There seems to be fields in the api to create users that were not in the ticket. is there a reason for this?
> > > Namely start date and ROLE?
> > 
> > 
> > Api fields correspond to the fields in the database, which include start date and role.
> 
> So should they be added to the modal?

Ultimately it's up to Stephen, my 2 cents is that role should be included, but start date you can either include or use `new Date()` when the submission button is clicked.

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+  { id: 'firstName', name: 'First Name', type: 'text' },
+  { id: 'surname', name: 'Surname', type: 'text' },
+  { id: 'jobTitle', name: 'Job title', type: 'text' },
+  { id: 'contractedHours', name: 'Contracted hours', type: 'number' },
+  { id: 'pay', name: 'Pay', type: 'number' },
+];
+
+const CreateUserModal = ({ closeModalFunction }: Props) => {
+  const dispatch = useDispatch();
+  const [prefferedDays, setPrefferedDays] = useState({
```
tjanowsk, line 37: typo in `preferred`

frontend/src/features/modalSlice.ts
```diff
@@ -0,0 +1,37 @@
...
+    showModal: (state, action: PayloadAction<ModalState>) => {
+      state.modalType = action.payload.modalType;
+      state.modalProps = action.payload.modalProps
+        ? action.payload.modalProps
+        : {};
+    },
+
+    hideModal: (state) => {
+      state.modalType = initialState.modalType;
+      state.modalProps = initialState.modalProps;
```
tjanowsk, line 28: Would `state = initialState` work? My understanding is that redux makes everything immutable, so it would create a copy of the initial state.
You may want to wrap initial state inside `Object.freeze` just to be sure.

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+import Button from '@material-ui/core/Button';
+
+import Clear from '@material-ui/icons/Clear';
+
+import 'toastr/build/toastr.min.css';
+import createUserModalStyle from './createUserModal.module.scss';
+
+interface Props {
+  closeModalFunction: () => void;
```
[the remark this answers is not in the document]
QuarkNerd: that's effectively what is passed in, I did it this way to make the specific modals don't need to know how they close and its handled in one place

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    Mon: false,
+    Tue: false,
+    Wed: false,
+    Thu: false,
+    Fri: false,
+    Sat: false,
+    Sun: false,
+  });
+
+  const [userInfo, setUserInfo] = useState<{ [key: string]: any }>({
```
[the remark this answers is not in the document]
QuarkNerd: I only added typing to avoid a strange error I was getting elsewhere, but I will do this too

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    firstName: '',
+    surname: '',
+    jobTitle: '',
+    contractedHhours: '',
+    pay: '',
```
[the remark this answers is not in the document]
QuarkNerd: good point

frontend/src/features/modalSlice.ts
QuarkNerd: It does not work. Because `state = initialState`, makes the local variable `state` point to `initialState` rather than change what `state` was initially pointing to. The redux state

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+              <TextField
+                id="outlined-basic"
+                label={field.name}
+                variant="outlined"
+                type={field.type}
+                value={userInfo[field.id]}
+                name={field.id}
+                onChange={handleUserInfoChange}
+              />
+            </span>
```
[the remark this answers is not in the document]
QuarkNerd: See this is interesting because Im used to keeping things together unless it calls for an entirely new component. So I wonder what others think?

frontend/src/components/employees/index.tsx
```diff
@@ -121,38 +119,32 @@ const renderTable = (tableColumns: TableColumn[]) => (data: StaffData[]) => (
...
-      onClick={addUser}>
-      Add
-    </Button>
-  </div>
-);
+const AddButton = () => {
+  const dispatch = useDispatch();
+
+  const createAddModal = () => {
+    dispatch(showModal({ modalType: 'CREATE_USER' }));
```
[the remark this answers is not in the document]
QuarkNerd: very good point

QuarkNerd:
image
someone please explain this error, I know that this is a string but apparently typescript does not

tjanowsk:
> image
> 
> image
> 
> someone please explain this error, I know that this is a string but apparently typescript does not

I think it's because name is an optional key, so it could be undefined.
I think `[event.target.name as string]` should work, but it will typecast undefined.

QuarkNerd:
> > image
> > image
> > someone please explain this error, I know that this is a string but apparently typescript does not
> 
> I think it's because name is an optional key, so it could be undefined.
> I think `[event.target.name as string]` should work, but it will typecast undefined.

image

tjanowsk:
> > > image
> > > image
> > > someone please explain this error, I know that this is a string but apparently typescript does not
> > 
> > 
> > I think it's because name is an optional key, so it could be undefined.
> > I think `[event.target.name as string]` should work, but it will typecast undefined.
> 
> image

Looks like typescrpt was OK with the key and now it's complaining about the value mismatching.
I think `event.target.value.toString()` will do it, we need to tell it how to get from `unknown` to `string | number`

QuarkNerd:
So there is certain info that is in this ticket but not in the definition of staffData. Are we intending this to change? If not, should I have 2 seperate interfaces for each type of staffData (shared base) or just use optional properties

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    dispatch(
+      createStaff({
+        firstName: userInfo.firstName as string,
+        surname: userInfo.surname as string,
+        jobTitle: userInfo.jobTitle as string,
+        role: userInfo.role as 'USER' | 'MANAGER',
+        contractedHours: userInfo.contractedHours as number,
+        pay: userInfo.pay as number,
+        preferredDates: prefDaysString,
+        startDate: toDateStr(new Date()),
```
tjanowsk, line 113: Do we need `toDateStr` at all?
We now have `date-fns` as a dependency, so we can just import `formatISO` and use
`formatISO(new Date(), { representation: 'date' })`
https://date-fns.org/v2.13.0/docs/formatISO

frontend/src/constants/global.ts
```diff
@@ -3,3 +3,26 @@ export const FULLTIME_HOURS = 37.5;
...
+    }
+    throw new Error(`Invalid date string: ${date}`);
+  } else {
+    const dateString = date.toISOString().split('T')[0];
+    if (checkValidDateStr(dateString)) {
+      return dateString;
+    }
+  }
+  throw new Error(`Shouldn't get here (invalid toDateStr provided): ${date}`);
+}
```
tjanowsk, line 28: Date and time libraries can handle this for you.
I imported `date-fns` in another ticket, so we can reuse it here.

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  const handleCreateClick = () => {
+    const prefDaysString = Object.values(preferredDays)
+      .map((preferred) => +preferred)
+      .join('');
+    dispatch(
+      createStaff({
+        firstName: userInfo.firstName as string,
+        surname: userInfo.surname as string,
+        jobTitle: userInfo.jobTitle as string,
+        role: userInfo.role as 'USER' | 'MANAGER',
```
tjanowsk, line 109: `'USER' | 'MANAGER'` is used a lot, I think this is a good use case for a string enum:
```
enum Role {
  USER = 'USER',
  MANAGER = 'MANAGER
}
```

Then use `Role` here, and replace `USER` with `Role.USER` and `'MANAGER'` with `Role.MANAGER`.

This comes with a few advantages:
- safety - in case of a typo, 'UESR' is a valid string, but not a valid role, so we want typescript to pick it up
- if we ever want to add another role, we'll just need to modify one enum.

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+    Mon: false,
+    Tue: false,
+    Wed: false,
+    Thu: false,
+    Fri: false,
+    Sat: false,
+    Sun: false,
+  });
+
+  const [userInfo, setUserInfo] = useState<{ [key: string]: string | number }>({
```
tjanowsk, line 51: Can we be more specific about the type and either reuse StaffData or create a similar object to hold the data?
At the moment we can set the state to an (almost) arbitrary object with string or number values.

QuarkNerd: Im gonna create a new type because staffdata doesnt quite fit

frontend/src/components/createUserModal/index.tsx
QuarkNerd: Good observation

frontend/src/constants/global.ts
QuarkNerd: Does date-fns allow us to do the typing of dates? Or do should I just replace the conversion/ check functions with it?

frontend/src/components/createUserModal/index.tsx
QuarkNerd: I did this because the way I defined the toDateStr function lets typescript know that what I am passing in is a string and at a date?

frontend/src/features/staffSlice.ts
```diff
@@ -42,6 +44,17 @@ export const fetchStaff = fetchWithAuth(
...
+export const createStaff = (staff: CreateStaffData): AppThunk => () =>
+  fetch(STAFF_CREATE_URL, {
+    method: 'POST',
+    headers: {
+      Authorization: getAuthenticationToken(),
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(staff),
+  });
+
```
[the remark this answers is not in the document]
emab: Yeah it will be slightly different. If this goes in before my PR I will just update it during the merge.

frontend/src/components/createUserModal/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+interface UserInfo {
+  firstName: string;
+  surname: string;
+  jobTitle: string;
+  email: string;
+  role: Role.USER;
+  contractedHours: number;
+  pay: number;
+  [key: string]: any;
```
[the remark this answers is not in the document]
QuarkNerd: its stops line 151 complaining, I dont know the proper way of handling this in typescript?

QuarkNerd:
> Looking better now, but I think the date handling can be made even simpler by replacing `toDateStr` with `formatISO`, after that I'll be happy to approve.

Im wondering if we can get a 2nd opinion on this, because im using that function for typescript purposes so that we can have a dateStr type

frontend/src/constants/global.ts
```diff
@@ -1,7 +1,32 @@
...
+
 export const URL = 'http://localhost:8080';
 export const FULLTIME_HOURS = 37.5;
 
 export const WINDOW_WIDTH_THRESHOLD: number = 780;
 
 // A placeholder for authentication token
 export const getAuthenticationToken = (): string => 'xx';
+
+enum DateStrBrand {}
```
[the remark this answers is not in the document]
QuarkNerd: https://spin.atomicobject.com/2017/06/19/strongly-typed-date-string-typescript/

I used this, im not sure if we actually want/need this but it felt like a good idea at the time

frontend/src/constants/global.ts
```diff
@@ -1,7 +1,32 @@
...
+
+export type DateStr = string & DateStrBrand;
+
+function checkValidDateStr(str: string): str is DateStr {
+  return isValid(parseISO(str));
+}
+
+export function toDateStr(date: Date | string): DateStr {
```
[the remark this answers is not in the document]
QuarkNerd: I made a design choice to have a strongly typed date string. I don't know if this is good or bad but it follows this https://spin.atomicobject.com/2017/06/19/strongly-typed-date-string-typescript/

Im assuming at somepoint the dates will be returned by the server as strings.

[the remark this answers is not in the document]
tjanowsk: The server will return strings inside JSON, but that can be handled by a separate function (e.g. `parseISO`).
I don't understand the objection to using `Date` objects just because they have time as well. We have a library that makes conversions and manipulations easy, I see little reason not to use `Date` everywhere. Maybe it's just me being lazy :stuck_out_tongue: 
But yeah, a second opinion would be nice.

PR #104, opened 2020-05-08
Feature/87 Roster - create basic view

tjanowsk:
rota


frontend/src/components/rota/rota.module.scss
```diff
@@ -1,5 +1,16 @@
 .rota {
   grid-area: main;
-  width: 100%;
-  height: 100%;
+  display: grid;
+  height: 100vh;
+  grid-template:
+    'header' min-content
+    'calendar' 1fr;
+  flex-direction: column;
```
AdrianBorg, line 8: is this relevant for `grid` display?
thought it's just for `flex`

frontend/src/components/rota/index.tsx
```diff
@@ -1,10 +1,58 @@
...
+  {
+    start: '2020-05-08T11:28Z',
+    end: '2020-05-08T13:33Z',
+    title: 'lunch',
+    description: 'FOOOD',
+    color: 'gold',
+  },
+];
+
+const tooltipFunction = (info: any) => {
```
AdrianBorg, line 28: is there a way of specifying what happens if the description is empty/undefined?
But actually, should we even worry about it?

tjanowsk: We'll be the ones choosing what to put in the description (I think it's name, surname, times and engagement type?), so I wouldn't worry about it at this stage.

frontend/src/components/rota/rota.module.scss
tjanowsk: Ah, I was playing around with different layouts and forgot to remove this. Will fix.

frontend/public/main.css
```diff
@@ -0,0 +1,7 @@
+body {
```
emab, line 1: Did this not work if it was in `index.module.scss`?

frontend/src/components/rota/rota.module.scss
```diff
@@ -1,5 +1,16 @@
...
+  margin: 0 1em;
+  grid-template:
+    'header' min-content
+    'calendar' 1fr;
+  h1 {
+    grid-area: header;
+  }
+}
+
+.calendarContainer {
```
AdrianBorg, line 14: > I changed the colour of 'today' header to yellow, but I couldn't do this in the rota's scss file, so I had to put it in main.css.
> My guess is that scss is compiled before fullcalendar.io classes are generated, so the styling is not applied to them. I have checked that the styling is not applied at all, it's not overwritten.
> There may be a way around this by e.g. playing with webpack's settings. Suggestions welcome.

would you have a similar result if instead you set the styling here with the global tag?

so something like:
``` 
.calendarContainer {
  grid-area: calendar;
  
  & :global(.fc-today) {
    background-color: yellow;
  }
}
```

Havent tried myself but I think that should make any child class of `calendarContianer` with the name `fc-today` get that styling.

tjanowsk: That worked, thanks!

PR #106, opened 2020-05-11
Feature/100 create login screen

emab:
## Changes
- Added a basic landing page
  - Displayed if the user isn't logged in. After login, user is redirected to the page they asked for.
  - Has some basic CSS using flexbox. I'm pretty new with flex box so would like to know if this is the correct way to centre stuff in this way?
- Added the landing page to the `App` component.
  - Added a some CSS to help display a loading screen. Loading is related to Auth0 authorisation stuff, usually is very small and shouldn't need to authenticate at any point again once logged in.
  - Loader may appear if the page is refreshed, but not if the user navigates through pages in the app.
- Removed the user email from `LoginOutButton` as it messed with the sidebar style.
  - Updated sidebar test accordingly.

### Screenshot
image


frontend/src/components/loginOutButton/index.tsx
```diff
@@ -1,17 +1,12 @@
 import React from 'react';
+import { Button } from '@material-ui/core';
```
tjanowsk, line 2: I thought we went for `import { Button } from '@material-ui/core/Button';` to keep the bundle size small, since we're not using babel?

emab: Woops! React did this automatically and I thought I'd changed all of these to import in that way.

Issue #109, opened 2020-05-12
BUG : Employee list capitalization


smangan360:
**Description**
As a tidy up task, each word in the employee list should have it's first letter capitalized. This will allow for uniformity throughout the employee list component. See the screenshot below:

![image](https://user-images.githubusercontent.com/54022972/81715296-f91e1380-946f-11ea-81db-30515f90e704.png)

Once validation takes place, the list should now be in this state:
![image](https://user-images.githubusercontent.com/54022972/81715573-4a2e0780-9470-11ea-9c7f-9abc24bfd4d3.png)

**Acceptance Criteria**: 
- Each entry in the full/part time field should begin with a capital letter
- Each job title should begin with a capital letter
- Names will be taken as the user enters them

QuarkNerd:
Do we not want to capitalize the 2nd word in name too?

PR #114, opened 2020-05-14
Capitalise names, job titles and Part/Full time

QuarkNerd:
#109 
Created a utility function to capitilse properly , using default export to make eslint happy


frontend/src/utils/string.ts
```diff
@@ -0,0 +1,4 @@
+const capitalizeFirstLetter = (s: string) =>
+  s.charAt(0).toUpperCase() + s.slice(1).toUpperCase();
```
tjanowsk, line 2: should this be `s.slice(1).toLowerCase()`?

QuarkNerd: yes it should be, good spot.

We can probably handle Sklodowska-Curie very easily by using the hypens. handling von Bismark requires either never changing name capitalization or having tons of exceptions. So this is one for the PO to weigh in on

PR #119, opened 2020-05-15
Bug/113 Add email field to user

emab:
Small update to include an email field in the staff object and DTO.

I've used hibernate to verify the email address in the DTO... do we need to do this? Will we be doing this check in the front-end?

### DIFF
--- a/backend/src/main/java/com/rota/api/dto/StaffDto.java
+++ b/backend/src/main/java/com/rota/api/dto/StaffDto.java
@@ -4,6 +4,7 @@ import com.rota.database.orm.staff.Role;
 import com.rota.database.orm.staff.Staff;
 import io.swagger.annotations.ApiModelProperty;
 import java.time.LocalDate;
+import javax.validation.constraints.Email;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotNull;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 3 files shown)


backend/src/main/java/com/rota/api/dto/StaffDto.java
```diff
@@ -44,6 +45,10 @@
   @ApiModelProperty("Is the staff currently inactive")
   boolean inactive;
 
+  @ApiModelProperty("The users email address")
+  @Email
```
emab, line 49: Do we need this verification here or can we assume the email coming in is good?

tjanowsk: I wouldn't assume anything about user input. If anything, assume that it can be malformed, malicious erc. :slightly_smiling_face:

backend/src/main/java/com/rota/api/dto/StaffDto.java
```diff
@@ -44,6 +45,11 @@
   @ApiModelProperty("Is the staff currently inactive")
   boolean inactive;
 
+  @ApiModelProperty("The users email address")
```
tjanowsk, line 48: Missing apostrophe in `user's` :scream:

emab: 😭

PR #125, opened 2020-05-18
Bug/121 multiple staff one email

emab:
### Changes
- Updated the `createStaff` method to check for a user with that email.
  - Throws a `DuplicateEmailException` with a descriptive message
- Updated `ControllerExceptionHandler` to send a `BAD REQUEST` when this exception occurs along with the descriptive message.
- Added `@Column(unique = true` to Staff object. This is an extra check to not allow any duplicate fields in email column, might be unnecessary?

_Many commits due to a fun rebase..._

### DIFF
--- a/backend/src/main/java/com/rota/api/StaffService.java
+++ b/backend/src/main/java/com/rota/api/StaffService.java
@@ -50,9 +50,10 @@ public class StaffService {
    * @return {@link Staff} object which has just been created.
    */
   public Staff createStaff(Staff newStaff) {
-    if (findStaffByEmail(newStaff.getEmail()).isPresent()) {
-      throw new DuplicateEmailException(newStaff.getEmail());
-    }
+    findStaffByEmail(newStaff.getEmail()).ifPresent(
+        (Staff staff) -> {
+          throw new DuplicateEmailException(newStaff.getEmail());
+        });
     return staffRepository.save(newStaff);
   }
 
--- a/backend/src/main/java/com/rota/api/StaffService.java
+++ b/backend/src/main/java/com/rota/api/StaffService.java
@@ -5,6 +5,7 @@ import com.rota.api.dto.StaffDto;
 import com.rota.database.orm.engagement.EngagementRepository;
 import com.rota.database.orm.staff.Staff;
 import com.rota.database.orm.staff.StaffRepository;
+import com.rota.exceptions.DuplicateEmailException;
 import com.rota.exceptions.StaffNotFoundException;
 import java.time.Instant;
 import java.util.List;
@@ -49,6 +50,9 @@ public class StaffService {
    * @return {@link Staff} object which has just been created.
    */
   public Staff createStaff(Staff newStaff) {
+    if (findStaffByEmail(newStaff.getEmail()).isPresent()) {
+      throw new DuplicateEmailException(newStaff.getEmail());
+    }
     return staffRepository.save(newStaff);
   }
 
--- a/backend/src/main/java/com/rota/database/orm/staff/Staff.java
+++ b/backend/src/main/java/com/rota/database/orm/staff/Staff.java
@@ -1,6 +1,7 @@
 package com.rota.database.orm.staff;
 
 import java.time.LocalDate;
+import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
@@ -44,6 +45,7 @@ public class Staff {
 
   private String jobTitle;
 
+  @Column(unique = true)
   private String email;
   
   @PrePersist
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -15,11 +15,6 @@ interface appState {
   targetUrl: string;
 }
 
-interface appState {
-  appState: Promise<RedirectLoginResult>;
-  targetUrl: string;
-}
-
 // Routes user to correct url after authenticating
 const onRedirectCallback = (appState: appState) => {
   history.push(
--- a/backend/src/main/java/com/rota/exceptions/ControllerExceptionHandler.java
+++ b/backend/src/main/java/com/rota/exceptions/ControllerExceptionHandler.java
@@ -1,6 +1,8 @@
 package com.rota.exceptions;
 
+import java.util.Objects;
 import javax.servlet.http.HttpServletRequest;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.ControllerAdvice;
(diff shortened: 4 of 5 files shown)


frontend/src/index.tsx
```diff
@@ -15,11 +15,6 @@ interface appState {
   targetUrl: string;
 }
```
emab, line 17: Not sure why this extra one was in here - was probably after a rebase.

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -49,6 +50,9 @@
    * @return {@link Staff} object which has just been created.
    */
   public Staff createStaff(Staff newStaff) {
+    if (findStaffByEmail(newStaff.getEmail()).isPresent()) {
+      throw new DuplicateEmailException(newStaff.getEmail());
+    }
```
tjanowsk, line 55: Instead of wrapping `isPresent` in an `if`, you can directly use `Optional`'s `ifPresent` function:
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#ifPresent-java.util.function.Consumer-
```
findStaffByEmail(newStaff.getEmail()).ifPresent(
  () -> throw new DuplicateEmailException(newStaff.getEmail())
);
```

I'm also not sure if this logic belongs here or in `StaffRepository.save`. Will we ever want to save a staff if there is duplicate e-mail?

backend/src/main/java/com/rota/database/orm/staff/Staff.java
```diff
@@ -44,6 +45,7 @@
 
   private String jobTitle;
 
+  @Column(unique = true)
```
tjanowsk, line 48: This should be enough to raise an exception if user wants to save a duplicate e-mail.
I'm wondering if it's enough to convert an SQL exception to DuplicateEmailException.
That way would save us one database lookup every time we add a new user.
Not a big deal either way, because I don't expect anyone to be spamming 'add user'.

backend/src/main/java/com/rota/api/StaffService.java
emab: This is how I thought I could use `ifPresent`, but I can't seem to get it to work? The code above doesn't compile for me anyway which I find weird. It's exactly how I wrote this initially and couldn't work out why it wasn't working.

> I'm also not sure if this logic belongs here or in StaffRepository.save. Will we ever want to save a staff if there is duplicate e-mail?

Very good point... currently if you update someone you won't get this error, just a JbdcViolationException or something along those lines.

backend/src/main/java/com/rota/database/orm/staff/Staff.java
emab: Initially I did this - I caught the ConstraintViolationException and threw a DuplicateEmailException but wasn't sure if that was a good idea, as other constraints may be broken and then we have a bad error message?

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: Oops, my bad, that's because `ifPresent` takes a `Consumer`, which is a function `Staff -> void`.
The following should work:
```
findStaffByEmail(newStaff.getEmail()).ifPresent(
  (Staff staff) -> {
    throw new DuplicateEmailException(newStaff.getEmail());
  }
);
```

backend/src/main/java/com/rota/database/orm/staff/Staff.java
tjanowsk: Good point. Currently it is our only constraint, but we shouldn't rely on that.

tjanowsk:
> _Many commits due to a fun rebase..._

Rebases are always fun aren't they :laughing: 
An interactive rebase might be able to help, i.e. run with the `-i` flag.

The commands would be (on your branch)
```
git fetch origin
git rebase -i origin/master
```

This will open your favourite text editor, which should be vim, and show all the commits preceded by `pick`.
You also get instructions on what else you can do, but you can simply delete all the lines corresponding to old commits and keeps only the ones for your PR (press `dd` to delete a line in vim, `ZZ` when done).
When done you can run the usual
`git push --force-with-lease`

Hope this helps!

backend/src/main/java/com/rota/api/StaffService.java
emab: Ahhh yeah that's it. I tried putting staff there but it didn't work, forgot that Java sometimes can infer the type. Had this problem before!

Updated anyway, thanks Tad! 👍

emab:
> > _Many commits due to a fun rebase..._
> 
> Rebases are always fun aren't they 😆
> An interactive rebase might be able to help, i.e. run with the `-i` flag.
> 
> The commands would be (on your branch)
> 
> ```
> git fetch origin
> git rebase -i origin/master
> ```
> 
> This will open your favourite text editor, which should be vim, and show all the commits preceded by `pick`.
> You also get instructions on what else you can do, but you can simply delete all the lines corresponding to old commits and keeps only the ones for your PR (press `dd` to delete a line in vim, `ZZ` when done).
> When done you can run the usual
> `git push --force-with-lease`
> 
> Hope this helps!

I really need to give this a go... problem is I didn't know which commits contained the stuff I really needed. Anyway it all turned out okay in the end! Next time I'll do this.

backend/src/main/java/com/rota/database/orm/staff/Staff.java
emab: image

This the current response when updating a staff member with an unavailable email currently. I could make the error handler check the message and decide if it is email related and send the correct exception? And if it can't decide what it is, throw another, less specific exception that we can dig into?

tjanowsk:
> I really need to give this a go... problem is I didn't know which commits contained the stuff I really needed. Anyway it all turned out okay in the end! Next time I'll do this.

My guess is that you want the last 4 commits, starting from 'Update to use access token email'?
Everything before that should already be in master.

backend/src/main/java/com/rota/database/orm/staff/Staff.java
tjanowsk: Someone more experienced should comment on this, you COULD search the string for `PUBLIC.STAFF(EMAIL)` for example, but this feels hacky and I'm worried it would make the code harder to read, so I'd be inclined to keep it as it is.

emab: Yeah I thought about this but wasn't sure if it was the best approach. We should leave it as it is and then when testers have a look we can decide what's best.

emab:
> Approved from me, if you want to try rebasing I can re-approve.
> If you do rebase, it may be wise to make a backup copy of the branch using
> `git branch my-backup-branch-name`
> but that's not necessary since everything gets squashed before merging to master.

Interactive re-base successful! 🎉

Issue #37, opened 2020-04-09
Get all shifts between dates endpoint


mangan360:
**Investigation needed on best way to do this. We may need two endpoints or one.**

As a manager I want to see the money being spent on wages for each job type of a period of a day, a week and a month. Each employee has an hourly rate attached to them, the endpoint should get all shifts worked. Breaks need to be deducted from each shift before displayed in the data. 

Acceptance Criteria: 
- Given when a manager queries the server for spending analysis for a single day then they should get back the spending analysis for each job sector (front of house, back of house, manager, ...). 
- Given when a manager queries the server for spending analysis for a week then they should get back spending analysis for each job sector (front of house, back of house, manager, ...). 
- Given when a manager queries the server for spending analysis for a month then they should get back the spending analysis for each job sector (front of house, back of house, manager, ...). 

**Notes**
- Breaks need to be deducted from the shifts
- Unpaid holidays should not be included
- Paid holidays should be included

emab:
> money being spent on wages for each job type

Currently job type is just a string which is typed by the manager, so we are relying on no spelling mistakes / consistency in their approach to naming job titles i.e chef != Chef != Cook != Kitchen. Is it outside MVP to ensure this data is consistent?

PR #131, opened 2020-05-19
Feature/37 - Get engagements between dates endpoint

tjanowsk:
I didn't realise this while creating the branch, but this is actually #37 .

This is needed for the pie chart.

Note that I have changed the endpoint to truncate the engagements if they cross the query time, e.g. if there is an engagement between 1pm-3pm and the query specifies the start time at 2pm, the returning engagement will be 2pm-3pm.
This will allow to calculate the times more precisely.

### DIFF
--- a/backend/src/main/java/com/rota/api/StaffService.java
+++ b/backend/src/main/java/com/rota/api/StaffService.java
@@ -36,7 +36,8 @@ public class StaffService {
     Instant startTime = start == null ? Instant.MIN : start;
     Instant endTime = end == null ? Instant.MAX : end;
 
-    return (Engagement engagement) -> !engagement.getEnd().isBefore(startTime)
+    return (Engagement engagement) -> !endTime.isBefore(startTime)
+        && !engagement.getEnd().isBefore(startTime)
         && !engagement.getStart().isAfter(endTime);
   }
 
@@ -61,7 +62,7 @@ public class StaffService {
 
   /**
    * Returns staff member's engagements between start and end dates (inclusive).
-   * Specifically engagements that start after the start date and end before the end date.
+   * Engagements that go through start or end times are truncated.
    * Both start and end dates are optional and can be left as null.
    *
    * @param staffId Staff member's ID
@@ -79,7 +80,7 @@ public class StaffService {
 
   /**
    * Returns all engagements between start and end dates (inclusive).
-   * Specifically engagements that start after the start date and end before the end date.
+   * Engagements that go through start or end times are truncated.
    * Both start and end dates are optional and can be left as null.
    * 
    * @param start start date or null to get from the beginning
--- a/backend/src/main/java/com/rota/api/StaffService.java
+++ b/backend/src/main/java/com/rota/api/StaffService.java
@@ -2,6 +2,7 @@ package com.rota.api;
 
 import com.rota.api.dto.EngagementDto;
 import com.rota.api.dto.StaffDto;
+import com.rota.database.orm.engagement.Engagement;
 import com.rota.database.orm.engagement.EngagementRepository;
 import com.rota.database.orm.staff.Staff;
 import com.rota.database.orm.staff.StaffRepository;
@@ -10,6 +11,8 @@ import com.rota.exceptions.StaffNotFoundException;
 import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -32,12 +69,26 @@
    * @return A list of member's engagements
    */
   public List<EngagementDto> getStaffEngagementsBetween(int staffId, Instant start, Instant end) {
-    Instant startTime = start == null ? Instant.MIN : start;
-    Instant endTime = end == null ? Instant.MAX : end;
     return engagementRepository.findByStaffId(staffId).stream()
```
emab, line 72: How do we think performance will be affected when someone has worked somewhere for a few years? Could be lots of engagements, will this cause issues? Just curious.

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -21,6 +24,40 @@
...
+
+    return (Engagement engagement) -> !engagement.getEnd().isBefore(startTime)
+        && !engagement.getStart().isAfter(endTime);
+  }
+
+  /**
+   * Truncates the engagement, e.g. if if starts before the start time the function will cut off the
+   * time difference and set the engagement start time to requested start time.
+   */
+  private Function<Engagement, Engagement> truncateEngagement(Instant start, Instant end) {
```
emab, line 46: I like how you've taken these out of the lambda expressions, I didn't know you could do this in Java.

backend/src/main/java/com/rota/exceptions/StaffNotFoundException.java
```diff
@@ -1,6 +1,8 @@
 package com.rota.exceptions;
 
 public class StaffNotFoundException extends RuntimeException {
+  static final long serialVersionUID = 1L;
```
emab, line 4: What does this do?

backend/src/test/java/com/rota/api/EngagementTests.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.time.Instant;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+public class EngagementTests {
```
emab, line 20: Good effort adding tests, I need to add some for methods I've created but I've been a bit lazy!

We could add tests for no engagements being returned and some negative tests such as invalid dates, start date after end date etc.

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -32,12 +69,26 @@
...
-                && !engagement.getEnd().isAfter(endTime))
+        .filter(isEngagementBetween(start, end))
+        .map(truncateEngagement(start, end))
+        .map(EngagementDto::fromEngagement)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Returns all engagements between start and end dates (inclusive).
+   * Specifically engagements that start after the start date and end before the end date.
```
emab, line 81: I think you changed this so that it gets engagements that overlap now as well?

backend/src/main/java/com/rota/exceptions/StaffNotFoundException.java
tjanowsk: I was getting a warning that a Serializable class needs a serialVersionUID, so this helps to suppress that.

Some explanation here
https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it#285809

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: Possibly, depends on company's retention policy for engagements.
Unfortunately, I can't think of another way of writing this at the moment.

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: Yes, forgot to update the comment

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: Everything in java is a class, `lambda`s are no exception.
Under the hood lambda expressions are classes from `java.util.function`.

Not sure if this is the most OO way of handling this, I may have been corrupted by functional programming.

backend/src/main/java/com/rota/exceptions/StaffNotFoundException.java
emab: Ah right didn't realise `RuntimeException` implemented `Serializable`!

Out of interest where did you see the warning?

backend/src/main/java/com/rota/api/StaffService.java
emab: I did some reading on it and some places were suggesting that a lambda shouldn't be too long otherwise it gets too complicated, so taking this logic out of there makes sense. If we needed to reuse `isEngagementBetween` in a different scenario a boolean might have been better, but we don't so I think this way is good!

backend/src/main/java/com/rota/api/StaffService.java
emab: Yeah neither can I.... I don't know enough about databases to know the performance hit.

The only think I can think is that the generated SQL for this will end up always getting all engagement records from the database and _then_ performing a filter. Is there a way to construct an SQL statement that would avoid this? It's probably not part of MVP to get this bit right though.

backend/src/main/java/com/rota/exceptions/StaffNotFoundException.java
tjanowsk: Neovim with coc.nvim plugin complains.
Given that this plugin is basically vscode's intellisense, I think vscode will complain too.

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: It should be possible to construct an sql statement which returns engagements between the dates.
I'll look into this, not sure if I'll make it part of this ticket though.

PR #136, opened 2020-05-20
Bug/129 Non-registered user can access app

emab:
Users could register and access the app if they weren't registered in the back-end. This PR addresses this by asking the front-end for the users role after logging in.

In the future, we might move to an invite only method.

### Changes 27/05/2020
- No longer use pageNotFound, instead if an invalid path is used the user is redirected to the `/summary` page.
- If the logged in user has `USER` role, they no longer see the `Employees` link in the sidebar.
  - Sidebar now takes `sidebarOptions` as props
  - Updated tests to check for this behaviour
- Added methods to `auth.ts` which check the user role/permissions.

### Changes
- Added `getRole` endpoint.
- Moved fullPageLoader to own component
- Updated Home component to render depending on Role

### Todo
Think about how to avoid an async call on page load. Somehow get the role into the access token? 

### Minor changes
- I changed an `Authentication` method name as it was an old name and no longer made sense.


frontend/src/components/home/index.tsx
```diff
@@ -1,32 +1,64 @@
...
+        setRole(json);
+      } catch (err) {
+        toastr.error('User not found');
+        setRole('NONE');
+      }
+    };
+    getRole();
+  }, [getTokenSilently]);
+
+  if (role === 'MANAGER' || role === 'USER') {
```
tjanowsk, line 34: Nothing wrong with using `if` here but the alternative is to use `switch`:
```
switch(role) {
  case Role.NONE:
    return foo;
  case Role.USER:
  case Role.MANAGER:
    return bar;
  default:
    return baz;
}
```
It may be more readable, but it might be just me.

frontend/src/components/home/index.tsx
```diff
@@ -1,32 +1,64 @@
...
+        <Sidebar />
+        <div className={`${homeStyles.right} ${homeStyles.column}`}>
+          <Header />
+          <main className={homeStyles.main}>
+            <Switch>
+              {role === 'MANAGER' && (
+                <Route path="/employees" component={Employees} />
+              )}
+              <Route path="/rota" component={Rota} />
+              <Route path="/summary" component={Summary} />
```
tjanowsk, line 46: Do we want `USER`s to see the summary screen?

frontend/src/components/home/index.tsx
```diff
@@ -1,32 +1,64 @@
...
+  if (role === 'MANAGER' || role === 'USER') {
+    return (
+      <div className={homeStyles.home}>
+        <Sidebar />
+        <div className={`${homeStyles.right} ${homeStyles.column}`}>
+          <Header />
+          <main className={homeStyles.main}>
+            <Switch>
+              {role === 'MANAGER' && (
+                <Route path="/employees" component={Employees} />
```
tjanowsk, line 43: Just a thought, but what happens if an employee manually tries to navigate to '/employees' endpoint?
We probably want to display some sort of 'unauthorised' message.
One way of doing this would be to have
`component={role === Role.MANAGER ? Employees : Unauthorised}`

As a side note, I'm worried that having `role === Role.MANAGER` may not scale well if we want to add more roles in the future. It may be better to have some `Role -> boolean` function `hasViewPermissions(role)`.
We did something similar on the back end.

frontend/src/components/home/index.tsx
emab: I think they want to see a summary screen, but potentially not the same summary screen as the managers. So potentially we could have the route `/summary` point to two different components if necessary.

frontend/src/components/home/index.tsx
emab: Yeah I think that is a better option, although if we move to use a method call as you suggested below to handle view permissions, an if statement might be better.

frontend/src/components/home/index.tsx
emab: Yeah definitely think this would be a better approach. I'll add some sort of basic unauthorised page.

And yes a method call would make more sense here. I've been told to look at react-router-guards for routing stuff too.

frontend/src/components/home/index.tsx
```diff
@@ -1,32 +1,70 @@
...
+
+  if (hasViewPermissions(role) && roleChecked) {
+    return (
+      <div className={homeStyles.home}>
+        <Sidebar />
+        <div className={`${homeStyles.right} ${homeStyles.column}`}>
+          <Header />
+          <main className={homeStyles.main}>
+            <Switch>
+              {role === 'MANAGER' && (
```
tjanowsk, line 43: Can we also use enum here?
`hasManagerPermissions(role)`

It may make sense to have several permission levels for different roles (i.e. a mapping `role -> number`), but that's probably post-MVP.

frontend/src/components/errorPage/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  <div className={style.fullPage}>
+    <div className={style.centered}>
+      <div>
+        <h1 className={style.header}>Error</h1>
+        <p>{error}</p>
+      </div>
+      <Divider orientation="vertical" flexItem />
+      <div>
+        <p>{message}</p>
+        {extra && (
```
tjanowsk, line 27: The `extra &&` may not be necessary, because if `extra` is undefined this will render <p><i></i></p>

frontend/src/components/errorPage/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+          <p>
+            <i>{extra}</i>
```
tjanowsk, line 29: Genuinely curious - what's the difference between the html `<i>` tag and CSS `font-style: italic;` ?

frontend/src/components/home/index.tsx
connorsettle: Ideally unauthorized users should not know anything about the admin side of the app. I mean if unauthorized user navigates to `/employees` they can just be redirected to a default page (in the same way they should if they navigate to `/invalid-path`.

Can be done with a guard like @emab said.

emab:
@tjanowsk @connorsettle what do you guys think about moving the current user role to the store rather than in the state? This is to allow the sidebar to decide what links to show. Or, I could pass the current role to the sidebar props?

frontend/src/components/home/index.tsx
emab: Currently they are navigated to a 404 style page, but not via router guards. It uses a router `switch` so if the page isn't recognised (which it won't be if they don't have the role) then it displays the page not found page. Is this suitable or is there a better way?

frontend/src/components/errorPage/index.tsx
emab: If extra is undefined then nothing renders, so this just makes sure the extra `<p>` tags aren't rendered if there is no extra message.

frontend/src/components/errorPage/index.tsx
emab: I'm honestly not sure, however from what I can tell most browsers just have a default CSS rule of `font-style: italic;` for this element.

frontend/src/service/auth.ts
```diff
@@ -0,0 +1,27 @@
...
+  const json = await response.json();
+
+  switch (json) {
+    case 'USER':
+      return Role.USER;
+    case 'MANAGER':
+      return Role.MANAGER;
+    default:
+      return Role.NONE;
+  }
```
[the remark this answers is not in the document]
emab: Cool that works nicely!

emab:
@connorsettle I looked into react router guards and it seemed like I would be adding more code and replacing the current `history` dependency if I used it. The behaviour we talked about is still achieved using the Routers `<Switch>` component and you are redirected to `/summary` if the page doesn't exist / not enough permissions.

Issue #93, opened 2020-05-06
Summary Screen: Create hourly breakdown chart


smangan360:
As a manager, I want to be able to see a total breakdown of employee working hours. This will allow for better understanding on expenses while giving a general breakdown of staff. The information should be represented in a pie chart. The total number of hours (including holidays) in that period should in the center of the pie chart and each time type should be allocated a colour. 

**Acceptance Criteria**
- There should be a pie chart on the top left of the summary page taking up about half the width of the page (see mock designs for an idea)
- There should be a header of the chart saying 'Hourly Breakdown'
- The chart should display the breakdown of hours for that selected period. (Including holidays)
- The chart should have the total numbers of hours worked in the center (similar to Kimble holiday pie chart) 
- Each job type should be given a specific colour. The colour breakdown will be displayed in the colour breakdown component. For testing purposes, include some text on each segment of the chart to show which job it represents. This can be removed later
- Sign into Kimble and observe how the pie chart loads. If possible it would be great to include a similar animation on load for our summary screen if possible. This is not a major requirement to include but it would be great to include. 

**Please see image below for styling**
![image](https://user-images.githubusercontent.com/54022972/81182325-8dd2ce00-8fa5-11ea-94ef-a40dc2632307.png)

**Notes**
- Tool tip is out of scope of this task

tjanowsk:
What are the job types? I think currently we only store engagementType which is either work or holiday.

PR #139, opened 2020-05-20
Feature/97 d3 chart with transitions

tjanowsk:
piechart

This PR to only includes the pie chart, #93 will be handled in a separate PR.


tjanowsk:
> This looks really good!
> 
> I don't know enough about d3 to comment on much of that code.

Just to explain what's happening:
To create a pie chart, I first use a `d3.pie` function, which turns the 'raw' data into pie chart data (start angle, end angle etc.). Then `d3.arc` converts this into and svg `path` element (it has some extra options like innerRadius, outerRadius, etc.).

The transitions are done using `.transition()` option, i.e. `initialState.transition().[what to change, how to do it etc]`. For the font, I simply start with font size 0 and then change the font size in the transition.
Transitioning pie chart arcs is more complicated and requires `.attrTween` on the path's `d` attribute (which defines the shape of the path. The argument of `attrTween` is a (borrowing language from topology) a homotopy function, which is a function that is a continuous interpolation between an initial and a final function.
Its signature is `data -> t -> string`, where `t` is a parameter that changes from `0` when the animation begins to `1` when it ends. In this case, I simply multiply start and end angles of all arcs by `t`, so they all start at `0` and finish where the data indicates.

Typescript makes this look more ugly than it needs to be with all the explicit typings and I'm not even sure if I'm using the types correctly.

Hope this clarifies things!

Issue #97, opened 2020-05-07
Summary Screen : D3 transitions for pie charts


smangan360:
**Description**
As a manager I want to be able be clearly see my employee hourly breakdown and holiday breakdown. A styling requirement is to have a transition on summary page load when rendering the pie charts. The functionality should mirror Kimble's pie chart load. 

**Acceptance Criteria**
- The two pie charts on the page should have a transition when they are being rendered
- It should mirror the functionality of Kimble

PR #140, opened 2020-05-25
Feature/93 create hourly breakdown chart

tjanowsk:
job_breakdown

Some questions:
- How should the time be calculated exactly? Only full hours? Up to a minute? Every started 15-min period? etc.
- How should holidays be displayed? A separate colour? A separate colour for each job type? Not at all?

This PR will require a rebase once get-engagements-between-endpoints and #97 are in master.


emab:
> * How should the time be calculated exactly? Only full hours? Up to a minute? Every started 15-min period? etc.

I guess it will be done in 15 minute blocks? I feel like setting that isn't much of a limitation when it comes to scheduling, but I guess that's something we need to decide on.

> * How should holidays be displayed? A separate colour? A separate colour for each job type? Not at all?

If it's paid holiday I guess you'd want it included to see where money is going. Should we have a checkbox to include holidays?

frontend/src/components/summary/PieChart.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { useEffect, useRef } from 'react';
+import * as d3 from 'd3';
+import tippy from 'tippy.js';
+import piechartStyle from './piechart.module.scss';
+
+interface JobTimes {
+  [key: string]: number;
+}
```
connorsettle, line 8: Can be generic (as with PieChartData). I would have 
```
export interface Dictionary<T> {
  [key: string]: T;
}
```
in somewhere like `model/common/dictionary.ts` (And have index.ts files throughout model folder).

frontend/src/components/summary/index.tsx
```diff
@@ -1,29 +1,86 @@
...
+  fetchEngagements,
+} from '../../features/engagementSlice';
+import { useAuth0 } from '../../react-auth0-spa';
+import { StaffData } from '../../constants/employees';
+import { EngagementData } from '../../constants/engagements';
+import capitalizeFirstLetter from '../../utils/string';
+
+const getStaffJob = (staffId: number) => (staffList: StaffData[]) => {
+  const staffMember = staffList.find((staff) => staff.id === staffId);
+  return staffMember ? staffMember.jobTitle : undefined;
```
connorsettle, line 20: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

Would mean result could be null or undefined but I guess you'd just check truthy anyway

tjanowsk: Nice, I didn't realise this was a thing!

Ideally I wanted to using something like an `Optional` in Java (sometimes called `Maybe`) and the closest thing to this I could find was https://folktale.origamitower.com/api/v2.3.0/en/folktale.maybe.html
but this package doesn't seem to go well with typescript :disappointed:

frontend/src/components/summary/PieChart.tsx
tjanowsk: This looks like such a common thing to use that I'm surprised that typescript doesn't provide this out of the box!

PR #148, opened 2020-06-01
Cleanup - remove dev/TokenButton

tjanowsk:
Replaced with a logger function that only prints to console in dev mode.


frontend/src/components/dev/logger.ts
```diff
@@ -0,0 +1,8 @@
+const logger = (description: string) => (arg: any) => {
+  if (process.env.NODE_ENV !== 'production') {
+    console.info(`${description}\n${arg}`);
```
connorsettle, line 3: Is arg necessarilly serializable to string? Could log the arg outside the string.

tjanowsk: Fixed!

PR #151, opened 2020-06-01
Cleanup/144 directory structure

tjanowsk:
Changed the directory structure, tried following 
https://redux.js.org/style-guide/style-guide#structure-files-as-feature-folders-or-ducks
in detailed explanation.

Not sure what a 'feature' is, it seems that this is there all the components go?
There is no one-to-one correspondence between tsx files and redux slices, so not sure how to handle this...


connorsettle:
@tjanowsk Please close in favor of #158

PR #157, opened 2020-06-03
Feature/155 - add engagement endpoint

tjanowsk:
Can add engament with
```
curl "http://localhost:8080/staff/addEngagement" -H "Authorization: BEARER $TOKEN" -X 'POST' -H 'Content-Type:application/json' -d '{"staffId":"1", "start":"2020-06-03T14:52:33Z", "end":"2020-06-03T17:00:00Z", "type":"SHIFT"}'
```
Verify with
```
curl "http://localhost:8080/staff/addEngagement" -H "Authorization: BEARER $TOKEN" -X 'POST' -H 'Content-Type:application/json' -d '{"staffId":"1", "start":"2020-06-03T14:52:33Z", "end":"2020-06-03T17:00:00Z", "type":"SHIFT"}'
```

where $TOKEN is set to the value shown in the front-end console.

### DIFF
--- a/backend/src/main/java/com/rota/api/StaffController.java
+++ b/backend/src/main/java/com/rota/api/StaffController.java
@@ -3,6 +3,7 @@ package com.rota.api;
 import com.rota.api.dto.EngagementDto;
 import com.rota.api.dto.StaffDto;
 import com.rota.auth.Authentication;
+import com.rota.database.orm.engagement.EngagementType;
 import com.rota.database.orm.staff.Role;
 import com.rota.database.orm.staff.Staff;
 import com.rota.exceptions.StaffNotFoundException;
@@ -167,5 +168,19 @@ public class StaffController {
   ) {
     staffService.removeStaff(id);
     return staffService.getActiveStaff();
+  }
+
+  /**
+   * Endpoint for the manager to add staff member's engagement.
+   */
+  @PostMapping("/staff/addEngagement")
+  @ApiOperation(value = "Allows manager to add an engagement for a certain user", authorizations = {
+      @Authorization(value = "Bearer") })
+  public void addEngagement(
+      @Valid 
+      @RequestBody 
+      EngagementDto engagement
+  ) {
+    staffService.addEngagement(engagement);
   }
 }
--- a/backend/src/main/java/com/rota/api/StaffService.java
+++ b/backend/src/main/java/com/rota/api/StaffService.java
@@ -95,6 +95,32 @@ public class StaffService {
         .collect(Collectors.toList());
   }
 
+  private Optional<Engagement> fromDto(EngagementDto engagement) {
+    Optional<Staff> staffOptional = staffRepository.findById(engagement.getStaffId());
+    return staffOptional.map((Staff staff) -> 
+    Engagement
+        .builder()
+        .staff(staff)
+        .start(engagement.getStart())
+        .end(engagement.getEnd())
+        .type(engagement.getType())
+        .build()
+    );
+
+  }
+
+  /**
+   * Adds an engagement to the database.
+   */
+  public void addEngagement(EngagementDto engagementDto) {
+    Optional<Engagement> engagementOptional = fromDto(engagementDto);
+    engagementOptional.ifPresentOrElse(
+        (Engagement engagement) -> engagementRepository.save(engagement),
+        () -> {
+          throw new StaffNotFoundException();
+        });
+  }
+
   /**
    * Adds a new staff entity to the database.
    *


backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -95,6 +95,32 @@
         .collect(Collectors.toList());
   }
 
+  private Optional<Engagement> fromDto(EngagementDto engagement) {
```
connorsettle, line 98: Would expect mapping function as part of EngagementDto class (at least this is recommended approach in C#)

tjanowsk: The slight complication here is that EngagementDto depends on StaffId, while Engagement depends directly on Staff object derived from the staffId, which needs to be fetched from the database (using staffRepository object).
I didn't want to introduce database dependencies to the DTO, so I put it here. The alternative is to inject staffRepository to another class, which I can do if you think it's better.

backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -168,4 +169,18 @@ public Staff updateStaff(
     staffService.removeStaff(id);
     return staffService.getActiveStaff();
   }
+
+  /**
+   * Endpoint for the manager to add staff member's engagement.
+   */
+  @PostMapping("/staff/addEngagement")
+  @ApiOperation(value = "Allows manager to add an engagement for a certain user", authorizations = {
+      @Authorization(value = "Bearer") })
```
connorsettle, line 178: How are we distinguishing manager / regular authenticated user?

tjanowsk: All `/staff` endpoints require manager authentication.

emab: I wonder if we should rethink the naming convention in the back-end too, and change `Staff` to `Employee` and clean up our endpoint naming or at least get it more consistent.

For example not all of our endpoints use verbs and some do, so for example with engagements we could use this pattern:

`GET /staff` returns all staff
`GET /staff/` returns staff with ID 1
`POST /staff` adds new staff object
`DELETE /staff/` removes staff object

`GET /staff/1/engagements` gets all staff 1's engagements
`POST /staff/1/engagements` adds an engagement to staff 1

and so on. What are peoples thoughts?

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -95,6 +95,32 @@
         .collect(Collectors.toList());
   }
```
emab, line 97: Should add some Javadoc comments here

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -95,6 +95,32 @@
...
 
+  private Optional<Engagement> fromDto(EngagementDto engagement) {
+    Optional<Staff> staffOptional = staffRepository.findById(engagement.getStaffId());
+    return staffOptional.map((Staff staff) ->
```
emab, line 100: What's the difference of using map vs `ifPresentOrElse`? If the staff member isn't found for example what happens here?

tjanowsk: https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#map-java.util.function.Function-

If the optional is empty, it will return empty optional, otherwise it will apply the function to the value.
Optional is java's attempt of introducing elements of functional programming and it's an example of a functor (a wrapper object with a map operation defined on it).

I used `map` here, because I want an Optional back `ifPresent` is a `void` method. They are analogous to `map` and `forEach` in javascript arrays.

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: It's a private function, so it won't show in Javadoc anyway (which is why checkstyle didn't enforce it).
I can add a comment if you think it will help understand the code.

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: We agreed to change /staff to /employee everywhere after the front-end cleanup PRs are merged.
Your proposed structure makes sense, the only wrinkle is that DELETE doesn't actually delete staff member from the database, so PUT might make more sense.

backend/src/main/java/com/rota/api/StaffService.java
emab: Oh yeah of course! Didn't notice that. Completely up to you.

backend/src/main/java/com/rota/api/StaffService.java
emab: Okay nice, so in this case if `staffOptional` is empty then an empty `Engagement` optional is returned. But if it is present, then it will perform the mapping and the Engagement builder returns an engagement optional? Or does it return an Engagement which gets turned into an optional automatically?

backend/src/main/java/com/rota/api/StaffController.java
emab: So when we set inactive we will treat it more as update to inactive I guess?

backend/src/main/java/com/rota/api/StaffService.java
tjanowsk: `Optional(x).map(f) = Optional(f(x))`
i.e. it takes the contents of the Optional, applies the function (which in this case builds Engagement) and wraps it in an Optional.

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: Addressed in #160

connorsettle: I agree with @emab I think.

Maybe need to review the api a little. If the employee entity cannot be retrieved after DELETE/PUT then it should be DELETE request.

Reason being that clients are not aware of underlying data or how api handles deleting and does not need to know. They don't care about inactive property.

PR #158, opened 2020-06-03
Refactor store and modal to get some better typings

connorsettle:


frontend/src/utils/date.ts
```diff
@@ -11,3 +11,6 @@ export const serializeDate = (date: Date = new Date()) => {
 
   throw new Error(`Invalid date: ${asString}`);
 };
+
+export const toDate = (value: Date | string) =>
```
tjanowsk, line 15: Do we need this function at all? Can't we handle Date and string inputs with two separate functions?
Also, we are using `date-fns` for date manipulations, so `parseISO`  or `parseJSON` can be used instead of `new Date(+value)`.
https://date-fns.org/v2.14.0/docs/parseJSON

frontend/src/components/rootModal/index.tsx
```diff
@@ -4,41 +4,46 @@ import { useDispatch, useSelector } from 'react-redux';
...
+/* eslint-disable react/jsx-props-no-spreading */
+const getModal = (modalState: ModalConfiguarion, defaultProps: ModalProps) => {
+  const modalType = modalState.type;
 
-const MODAL_COMPONENTS: ModalHash = {
-  DELETE_USER: DeleteUserModal,
-  CREATE_USER: CreateUserModal,
+  switch (modalState.type) {
+    case ModalType.DeleteUser: {
+      const { type, ...rest } = modalState;
```
tjanowsk: Minor nitpick, I believe this line can be pulled out of the switch statement (can replace `const modalType = modalState.type`)

frontend/src/store/engagementSlice.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const fetchEngagements = (
+  start: Date,
+  end: Date,
+  token?: string
+): AppThunk => (dispatch) =>
+  get(engagementsFetchUrl(start, end), token)
+    .then((response) => response.json())
+    .then((response) => dispatch(setEngagements(response)))
+    .catch((err) => toastr.error(err));
```
tjanowsk, line 42: I know this is unchanged from before and not too important either, but this line can be shortened to
`.catch(toastr.error)`
I can't find a linting rule that would do that though. Either way, I'm fine with this staying the way it is.

frontend/src/store/modalSlice.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+interface CreateUserModalConfigaration extends CreateUserConfigaration {
+  type: ModalType.CreateUser;
+}
+
+export type ModalConfiguarion =
+  | DeleteUserModalConfigaration
+  | CreateUserModalConfigaration;
+
```
emab, line 23: typo on configuration in this file

frontend/src/components/createUserModal/index.tsx
```diff
@@ -12,15 +12,14 @@ import Clear from '@material-ui/icons/Clear';
...
+import { ModalProps, RoleType } from '../../model';
+import { createEmployee } from '../../store/employeeSlice';
 import createUserModalStyle from './createUserModal.module.scss';
 
-interface Props {
-  closeModalFunction: () => void;
-}
+export interface CreateUserConfigaration {}
+
+type Props = CreateUserConfigaration & ModalProps;
```
emab, line 23: typo on configuration

frontend/src/components/RootModal/index.tsx
```diff
@@ -0,0 +1,52 @@
...
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import CreateUserModal from '../CreateUserModal';
+import DeleteUserModal from '../DeleteUserModal';
+import { ModalProps } from '../../model';
+import {
+  hideModal,
+  ModalConfiguarion,
+  ModalType,
```
emab, line 11: config typo in this file

frontend/src/components/rootModal/index.tsx
connorsettle: Afraid its not possible https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions

If we destructure the `type` & `rest` outside the switch the variables become disassociated and we don't know what type `rest` is. 

The reason for `modalType` var is because the default case is strictly currently never possible and TS knows this. I added the default case for when we add a new `ModalType` and forget to add case to the switch statement.

frontend/src/components/RootModal/index.tsx
connorsettle: Oops, I will add spell checker extension :facepalm:

frontend/src/utils/date.ts
connorsettle: Reason is that I didn't want to add serial interface for state entity. Would rather just parse the object without knowing if it been serialized or not from selector perspective :)

frontend/src/components/rota/index.tsx
```diff
@@ -27,7 +28,7 @@ const DUMMY_EVENTS = [
   },
 ];
 
-const tooltipFunction = (info: any) => {
+const tooltipFunction = (info: { event: EventApi; el: HTMLElement }) => {
```
tjanowsk, line 31: Nice! Is there an easy way of looking up these types? I tried looking for them but settled for `any` in the end...

connorsettle: I just got those types from VSC. Unfortunatley `info` itself is not a named typed. We'd need to extract from the function so for me its cleaner to just take the parts we need.

image

connorsettle:
> github checks freeze when running `npm run build --if-present`, do we know why?

Issue was github not recognising file name changes that are just case changes. You can look it up. It was fixed by using `git mv ...` rather than file/folder name changes

PR #161, opened 2020-06-08
fix the front-end endpoint names to match the back-end

tjanowsk:
Making the front-end work with back-end again.

NB. I couldn't use `delete` for the name of the delete function, because that seems to be a keyword.


frontend/src/services/apiService.ts
```diff
@@ -9,6 +9,12 @@ export const put = async (request: RequestInfo, token?: string) =>
     headers: { Authorization: `Bearer ${token}` },
   });
 
+export const del = async (request: RequestInfo, token?: string) =>
```
connorsettle, line 12: `remove`/`deleteEntity`?

PR #164, opened 2020-06-08
WIP - Feature 88 - Fetch and display engagement data on rota screen

tjanowsk:
rota

Stuff to do:

- currently all engagements are fetched, limit the engagements to currently viewed week
- only the manager part of the view works, we need to decide how to best handle the user view (e.g. use different component or check permissions within the rota component)
- decide what to do with the colours, currently hard-coded as a colour array and cycle when the number of users is greater than the number of colours - perhaps there is a better solution?
- The text in rendered boxes is cut off - decide how we want the text to be displayed if there are multiple user engagements at the same time. Perhaps a different calendar view would be more appropriate (but timeline-view https://fullcalendar.io/docs/timeline-view requires premium version of fullcalendar.io)


connorsettle:
Looks good

frontend/src/store/engagementSlice.ts
```diff
@@ -11,15 +11,19 @@ import { get } from '../services/apiService';
...
 const engagementAdapter = createEntityAdapter<Engagement>({
   selectId: ({ staffId, start }) => `${staffId}::${start}`,
 });
 
-const engagementsFetchUrl = (start: Date, end: Date) =>
-  `${baseUrl}/staff/shifts?start=${start.toISOString()}&end=${end.toISOString()}`;
+const engagementsFetchUrl = async (start: Date, end: Date, token?: string) =>
+  isManagerRole(await getRole(token))
+    ? `${baseUrl}/staff/shifts?start=${start.toISOString()}&end=${end.toISOString()}`
+    : `${baseUrl}/myShifts?start=${start.toISOString()}&end=${end.toISOString()}`;
```
connorsettle, line 26: For me it could be the same endpoint. But its not something that needs changed now.

PR #165, opened 2020-06-09
Feature/137 add engagement modal

tjanowsk:
create-engagement-modal

Question (possibly post-MVP): how to handle repeated events?


frontend/src/components/CreateEngagementModal/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+
+  const handleEmployeeChange = (
+    event: React.ChangeEvent<{ name?: string | undefined; value: unknown }>
+  ) => {
+    setEmployeeId(event.target.value as number);
+  };
+
+  const handleEngagementTypeChange = (
+    event: React.ChangeEvent<{ name?: string | undefined; value: unknown }>
```
connorsettle, line 40: `name?: string | undefined` - seems like either `?` or `| undefined` is redundant. Also name is never used so might not be required at all.

Is value the type of `event.target.value`? If so might as well give it expected type.

```suggestion
    event: React.ChangeEvent<{ value: EngagementType }>
```

PR #168, opened 2020-06-12
Feature/134 change delete functionality

emab:
### Changes
- Updated the `DELETE` endpoint in the controller to use a RESTful url
  - To delete user with ID of 1, you use `DELETE /staff/1`
  - This has been updated in the front-end.
- Since email is set to `null` on staff deletion, an optional is used for `getEmail()`.

### DIFF
--- a/backend/src/main/java/com/rota/api/StaffController.java
+++ b/backend/src/main/java/com/rota/api/StaffController.java
@@ -18,6 +18,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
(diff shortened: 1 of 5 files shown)


backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -154,11 +155,11 @@ public StaffDto updateStaff(
    * @param id staff id to remove.
    * @return An updated list of active staff members.
    */
-  @DeleteMapping("/staff")
+  @DeleteMapping("/staff/{id}")
```
connorsettle, line 158: :+1:

backend/src/main/java/com/rota/api/StaffService.java
```diff
@@ -169,8 +172,7 @@ public void removeStaff(int id) {
   public Optional<Staff> findStaffByEmail(String email) {
     return staffRepository.findAll().stream()
         .filter(staff -> staff
-            .getEmail()
-            .equalsIgnoreCase(email))
+            .getEmail().isPresent() && staff.getEmail().get().equalsIgnoreCase(email))
```
tjanowsk, line 175: this can also be written as
`staff.getEmail().filter(email::equalsIgnoreCase).isPresent()`
which saves one `getEmail` call. Not a big deal overall.
Filter on an optional will return the original optional if the predicate is true or an empty optional otherwise.

backend/src/main/java/com/rota/api/StaffController.java
tjanowsk: Just curious - are there any guidelines on the use of path variables vs query parameters?

emab: I don't think there are guidelines, but the general practice that I've seen is that you use path variables for accessing resources, and use query parameters to filter and sort those resources.

backend/src/main/java/com/rota/api/StaffService.java
emab: No I like your way! I was trying to work out how to do this but ended up doing it this other way as a fallback. I like the functional filter chaining!

PR #172, opened 2020-06-17
Feature/143 edit staff

### LINKED ISSUE #143: Add option to edit user
There are 2 main possibilities:
1. use a modal (basically re-purpose the 'add user' modal)
1. use ag-grid's edit functionality https://www.ag-grid.com/javascript-grid-cell-editing/

If we go with the latter option, we will have to restructure the table, which is currently NOT in one-to-one correspondence with the data.
For example 'name' column is squashed 'firstName' and 'surname', 'Part/full time' column is derived from 'workingHours' (with a loss of information) and some columns are missing altogrther (start date, salary etc.)

tjanowsk:
Added option to edit staff.

Also cleaned up the employee modal (removed unnecessary types etc.)

Changed the way POST endpoint for the employees works to return a full updated list of employees for convenience.

closes #143


backend/src/main/java/com/rota/api/StaffController.java
```diff
@@ -137,12 +137,12 @@ public Role getRole() {
    * Endpoint to update the information of a staff member.
    *
    * @param updatedStaff Updated staff information.
-   * @return The updated {@link Staff} member.
+   * @return The updated {@link Staff} list.
```
emab, line 140: I think this way is fine for now as we will not have many employees, but returning the updated staff member and replacing the one in the state with the updated version from the server is definitely the way to do it if we were dealing with more.

frontend/src/components/CreateUserModal/index.tsx
```diff
@@ -44,40 +54,39 @@ const textFields: TextField[] = [
...
+const UserModal = ({ onClose, currentEmployee, SubmitButton }: Props) => {
+  const [userInfo, setUserInfo] = useState<Employee>(currentEmployee);
 
   const handlePreferredDaysChange = (
     event: React.ChangeEvent<HTMLInputElement>
   ) => {
-    setPreferredDays({
-      ...preferredDays,
-      [event.target.name]: event.target.checked,
+    const weekday = +event.target.name;
```
connorsettle, line 81: weekday must be an index? Maybe `dayIndex` is better

frontend/src/components/CreateUserModal/index.tsx
```diff
@@ -44,40 +54,39 @@ const textFields: TextField[] = [
...
+  id: 0,
+  firstName: '',
+  surname: '',
+  jobTitle: '',
+  email: '',
+  role: RoleType.User,
+  contractedHours: 0,
+  hourlyRate: 0,
+  startDate: new Date(),
+  preferredDates: '0000000',
```
connorsettle, line 72: For me its not the clearest representation of weekdays.

tjanowsk: I haven't changed this since before - this is how it's stored in the database (we couldn't agree on the best way of storing this and we didn't want to create  a separate table for preferred dates).
We could use an object `{[weekday: string]: boolean}` on the front-end, but we'll have to think how to serialize it and then process it on the back end.
This could possibly be a separate ticket?

connorsettle: Ok

PR #174, opened 2020-06-18
Feature/135 display user name on login

emab:
User nickname is now displayed in header.

- Added User type interface to `auth0Spa` as it was `unknown` before.

image

### DIFF
--- a/frontend/src/components/Header/header.module.scss
+++ b/frontend/src/components/Header/header.module.scss
@@ -5,9 +5,17 @@
   background-color: $colour-dark-blue;
   display: grid;
   grid-template-areas: 'left middle right';
-  grid-template-columns: 2.5rem 1fr 100px;
+  grid-template-columns: 2.5rem 1fr;
 }
 
 .headerRight {
+  display: flex;
   grid-area: right;
+}
+
+.userName {
+  color: $colour-text-primary;
+  padding-right: 10px;
+  position: relative;
+  bottom: 5px;
 }
--- a/frontend/src/components/Header/header.module.scss
+++ b/frontend/src/components/Header/header.module.scss
@@ -10,12 +10,11 @@
 
 .headerRight {
   display: flex;
+  align-items: center;
   grid-area: right;
 }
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/Header/header.module.scss
```diff
@@ -5,9 +5,17 @@
...
-  grid-template-columns: 2.5rem 1fr 100px;
+  grid-template-columns: 2.5rem 1fr;
 }
 
 .headerRight {
+  display: flex;
   grid-area: right;
 }
+
+.userName {
```
emab, line 16: This might not be the best way to do this - @tjanowsk any ideas on how to get the username text to be centered in the header?

tjanowsk: I seem to recall that setting left and right margins to auto can be used to center the text.
Alternatively, can try adding `align-content` and `justify-content` to `.headerRight`
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Aligning_Items_in_a_Flex_Container

frontend/src/components/Header/index.tsx
```diff
@@ -2,13 +2,18 @@ import React from 'react';
...
-      <LoginOutButton />
-    </div>
-  </header>
-);
+const Header = () => {
+  const { user } = useAuth0();
+  return (
+    <header className={headerStyles.headerContent}>
+      <div className={headerStyles.headerRight}>
+        <p className={headerStyles.userName}>{user && user.nickname}</p>
```
tjanowsk, line 12: Maybe `<span>` would be more appropriate than `<p>`?

frontend/src/components/Header/header.module.scss
emab: Yeah I tried `align-items`, and it moved the login button down level with the text. I'll have another look.

emab:
Changing `<p>` to `<span>` helped with the CSS issues - `align-items: centre` now works as expected.

PR #178, opened 2020-06-19
Deployment/heroku

emab:
- Split application properties
  - Shared properties stay in `application.properties`
  - Dev properties are in `application-dev.properties`
  - Prod properties are in `application-prod.properties`
  - `gradlew bootRun` defaults to using dev properties
    - Prod properties can be run using `gradlew bootRun -Pprofile=prod`
- Added mysql connector

### DIFF
--- a/backend/src/main/resources/application-dev.properties
+++ b/backend/src/main/resources/application-dev.properties
@@ -1,0 +1,7 @@
+spring.datasource.url=jdbc:h2:mem:testdb
+spring.datasource.driverClassName=org.h2.Driver
+spring.datasource.username=user
+spring.datasource.password=password
+spring.jpa.database-platform=org.hibernate.dialect.H2Dialect
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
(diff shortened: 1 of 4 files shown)


backend/src/main/resources/application-prod.properties
```diff
@@ -0,0 +1,5 @@
+spring.datasource.url=jdbc:${CLEARDB_DATABASE_URL}
+spring.jpa.database-platform=org.hibernate.dialect.MySQL5Dialect
+spring.jpa.hibernate.ddl-auto=create-drop
+spring.security.oauth2.resourceserver.jwt.issuer-uri = https://rota-flex-101.eu.auth0.com/
+auth0.audience = rota-flex-api
```
tjanowsk, line 5: These are repeated in application.properties. Is this intentional?

backend/src/main/resources/application-dev.properties
```diff
@@ -0,0 +1,7 @@
+spring.datasource.url=jdbc:h2:mem:testdb
+spring.datasource.driverClassName=org.h2.Driver
+spring.datasource.username=user
+spring.datasource.password=password
```
tjanowsk, line 4: This may be OK for dev, but we should avoid committing passwords in general.
How is the production database authentication handled?

backend/src/main/resources/application-prod.properties
emab: Not intentional!

backend/src/main/resources/application-dev.properties
emab: The production passwords are stored in the config evironment in heroku. They are build into the msql url, so mysql:<username>:<password>@<dbaddress>.

PR #184, opened 2020-06-22
Feature/162 implement axios

### LINKED ISSUE #162: Use Axios for network requests
Reasons:
1. Support for cross browsers
2. Interceptor to set the bearer token rather than passing through state.

https://github.com/axios/axios

tjanowsk:
Replaced fetch with axios.

Also fixed a bug where an engagement could not be edited/removed before refreshing the screen first.

closes #162

### DIFF
--- a/frontend/src/services/apiService.ts
+++ b/frontend/src/services/apiService.ts
@@ -1,18 +1,29 @@
+import axios from 'axios';
+import { environment } from '../utils/environment';
+
 const headers = (token?: string) => ({
   Authorization: `Bearer ${token}`,
   'Content-Type': 'application/json',
 });
 
-const apiRequest = (method: string) => async (
-  request: RequestInfo,
+const axiosInstance = axios.create({
+  baseURL: environment.baseUrl,
+  timeout: 1000,
+});
+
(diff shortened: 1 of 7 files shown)


frontend/src/services/apiService.ts
```diff
@@ -1,18 +1,29 @@
...
-const apiRequest = (method: string) => async (
-  request: RequestInfo,
+const axiosInstance = axios.create({
+  baseURL: environment.baseUrl,
+  timeout: 1000,
+});
+
+const apiRequest = (method: 'GET' | 'PUT' | 'POST' | 'DELETE') => async (
+  url: string,
   token?: string,
```
connorsettle, line 16: is it possible we can set the token via interceptor?

connorsettle: Not sure how straightforward it will be with current auth0 implementation

tjanowsk: I think useAuth0 has to be wrapped in a react component at the moment, I'll see if there is a workaround.

tjanowsk: This seems like a fairly big task and more than one way of going about this so I opened #185

PR #188, opened 2020-06-24
Feature/185 move token to axios interceptor

### LINKED ISSUE #185: Move token authentication to axios interceptor
Currently we are using useAuth0 from the react wrapper, which can only be used inside  React components.

There are two ways around this:
- replace react auth0 with plain javascript one
- create a new React context that will pass get, set, put and post functions tochild components https://reactjs.org/docs/context.html

@emab @connorsettle  any thoughts?

tjanowsk:
A hack which makes auth0 config available as both a global variable and react component

closes #185

### DIFF
--- a/frontend/src/services/apiService.ts
+++ b/frontend/src/services/apiService.ts
@@ -1,5 +1,6 @@
 import axios from 'axios';
 import { environment } from '../utils/environment';
+import { authWrapper } from '../auth0Spa';
 
 const headers = (token?: string) => ({
   Authorization: `Bearer ${token}`,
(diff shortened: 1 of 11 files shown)


frontend/src/services/apiService.ts
```diff
@@ -13,14 +14,13 @@ const axiosInstance = axios.create({
...
   url: string,
-  token?: string,
   data?: object
 ) =>
   axiosInstance
     .request({
       url,
       method,
-      headers: headers(token),
+      headers: headers(await authWrapper.authClient?.getTokenSilently()),
```
connorsettle, line 23: Shame its only available like this :(

Still nice to not have to pass accessToken all through business/app logic :+1:

